PR #111, opened 2011-04-02
Delete Content-Type header from HttpAccepted (204 No Content) responses

kmike:
This header can be returned with HttpAccepted according to rfc2616, but it is useless (so removing it will slightly decrease bandwidth usage) and lint tools can complain about this header (e.g. Ian Bicking's wsgi lint middleware included in WebTest raises AssertionError for 204 responses with Content-Type header).


SeanHayes:
Superceded by #1054.

PR #155, opened 2011-06-02
Use setuptools in setup.py

claymation:
This one-line change prevents an ``install_requires`` warning when running ``python setup.py install`` and adds support for running ``python setup.py develop``.


SeanHayes:
Fixed

PR #176, opened 2011-06-23
allow both classes and instances of Resource to be registered in Api

michaelmwu:
Since model ForeignKeys and ManyToManyFields use classes as the key, and so do resource ToOneField and ToManyField, both options would be nice. No reason to type the extra () to construct an instance.


SeanHayes:
I'm not sure it's worth the extra complication to our code.

PR #191, opened 2011-07-06
Generic relations

dhatch:
I've added support for Generic Relations in the API by modifying the RelatedField to allow the ```to``` field to be a dictionary which maps django models to Resources.  Basic usage is as follows:

Django Model:

```python
class GenericTag(models.Model):
    name = models.CharField(max_length = 30)
    object_id = models.PositiveIntegerField()
    content_type = models.ForeignKey(generic.ContentType)
    content_object = generic.GenericForeignKey()
    
    def __unicode__(self):
        return u"%s" % (self.name)
```

Resource:

```python
class GenericTagResource(ModelResource):
    content_type = fields.ContentTypeField()
    content_object = fields.ToOneField(
        {Category : CategoryResource, Taggable : TaggableResource}, attribute="content_object", contenttype_field=content_type)

    class Meta:
        resource_name = 'generictag'
        queryset = GenericTag.objects.all()
        authorization = Authorization()
```

The ```to``` field dictionary tells the ```RelatedField``` class what Resources to use when hydrating/dehydrating.  A ```contenttype_field``` reference must also be passed if writing is allowed which allows for the proper type object to be written.  Two new classes were added, ```ContentTypeField``` which provides a shortcut for this purpose and ```ContentTypeResource``` which users can subclass to customize.  I've left it up to users to add ```ContentTypeResource``` to their urlconf by using ```Api.register``` or some other method.  All functionality is supported including POST/PUT with nested resources.  I've also written tests for the code in the ```GenericForeignKeyTest``` class inside ```tests/related_resource/tests.py``` The reverse ```GenericRelation``` is supported by using the ```ToManyField```.  All tests pass when run with ```tests/run_all_tests.sh```

### DIFF
--- a/tastypie/fields.py
+++ b/tastypie/fields.py
@@ -440,7 +440,7 @@ class RelatedField(ApiField):
         if self.contenttype_field and not isinstance(self.to, dict):
             raise ValueError(
                 "to argument must be a dictionary " + 
-                "when used with contenttype_attribute")
+                "when used with contenttype_field")
 
         if self.contenttype_field and not issubclass(type(self.contenttype_field), ToOneField):
             raise ValueError(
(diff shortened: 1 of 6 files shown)


SeanHayes:
I believe this functionality is now in tastypie.contrib.contenttypes.

PR #227, opened 2011-08-14
Add verbose_name to schema

cyberj:
This patch adds `verbose_name` from django's models.Field or Tastypie Field into API's `get_schema` view.

It can be usefull to build automatics forms with this API schema with a more verbose name than resource's field name.

Hope it can be usefull.


aleksey-kutepov:
+1

SeanHayes:
Fixed by #1370. Thanks!

PR #229, opened 2011-08-18
added sphinx documentation directive

imosquera:
I've added a sphinx documentation directive so you can add sphinx like documentation.  By adding this simple line:

.. tastydoc:: project.application.urls.v1_api

that will document the entire API with endpoints in a sphinx like fashion.

### DIFF
--- a/tastypie/resources.py
+++ b/tastypie/resources.py
@@ -404,7 +404,10 @@ class Resource(object):
             verbose = verbose.lower() == 'true'
             for field_name, field_object in self.fields.items():
                 if isinstance(field_object, ForeignKey):
-                    field_object.full = verbose
+                    if verbose:
+                        field_object.full = verbose
+                    else:
+                        del(self.fields[field_name])
                     
     def dispatch(self, request_type, request, **kwargs):
         """


tastypie/resources.py
```diff
@@ -397,6 +398,17 @@ def dispatch_detail(self, request, **kwargs):
         """
         return self.dispatch('detail', request, **kwargs)
     
+    def set_verbosity(self, request):
```
SeanHayes: I think this was included by accident.

tastypie/resources.py
```diff
@@ -397,6 +398,17 @@ def dispatch_detail(self, request, **kwargs):
...
+        verbose = request.GET.get('verbose', None)
+        if verbose is not None:
+            verbose = verbose.lower() == 'true'
+            for field_name, field_object in self.fields.items():
+                if isinstance(field_object, ForeignKey):
+                    if verbose:
+                        field_object.full = verbose
+                    else:
+                        del(self.fields[field_name])
```
SeanHayes: Field objects are reused between requests, you shouldn't modify or delete them.

PR #235, opened 2011-08-24
Fix to allow API keys to be used in test fixtures

jpuderer:
Fix to allow API keys to be prepopulated in test fixtures, when the create_api_key signal hook is used.


SeanHayes:
Was duplicated by #305

PR #243, opened 2011-08-31
ApiAccess.url was to small for urls with a lot of GET params; 

exister:
Migration to TextField added;
Fixed misspell in ApiAccess.unicode() method;


georgedorn:
Re-implemented in #1522

PR #254, opened 2011-09-17
When the api receives requests without a content type, the server responds with a 500.  This patch allows for a default content type to be set in the settings file. 

johncosta:
Allows a default content type to prevent 500 errors. Unit tests included.

[Review on Reviewable](https://reviewable.io/reviews/django-tastypie/django-tastypie/254)


SeanHayes:
I think this issue has been fixed, we now default to JSON all over the place.

PR #266, opened 2011-10-17
pagination using HTTP_RANGE

william-gr:
A few REST clients use HTTP_RANGE from RFC to paginate results, instead of GET parameters...

I've made a few changes to make it happen if tastypie...

The changeset 4f5028e7  
 was not supposed to be in this pull request but i don't know how to cherry-pick (pull request) in github..

I've tested it with Dojo toolkit + datagrid and seems to behave properly.. I'm not familiar with tastypie codebase so I don't really know if thats the best approach to implement this (it should not change the default pagination behavior)

Thanks...


william-gr:
Guess this is not going to get merged after 10 years.

Issue #199, opened 2011-07-19
validation.FormValidation cannot validate model update with unique field.


overminder:
Tastypie version: 1.0.0 beta
Django version: 1.3
Python version: 2.7.1
Platform: Ubuntu 11.04

My modelresource uses a tastypie.FormValidation instance created by a django ModelForm as its validator.
However, my model contains a unique field and thus when using HTTP PUT to update an existing model instance, the validator will report error on duplicate field values.

According to [a SO post](http://stackoverflow.com/questions/526457/django-form-fails-validation-on-a-unique-field), when doing model update, the form should be instantiated by an instance of the model that is going to be updated.

A dirty patch is attached but I still haven't figured out how to fix this in a better way...
# tastypie/validation.py:52
```
if request.method == 'PUT' and 'id' in bundle.data:
    # if we are updating a field, the model instance must also be
    # included in the arguments.
    instance = bundle.obj.__class__.objects.get(pk=bundle.data['id'])
    form = self.form_class(data, instance=instance)
else:
    form = self.form_class(data)
```

SeanHayes:
Should be fixed by c5451b90b18b0cb64841b2276d543230d5f58231

PR #270, opened 2011-10-20
Moving validation cycle after full_hydrate.

theorm:
``FormValidation`` right now is unable to access a hydrated object. However it is needed when a form is trying to validate an updated existing object with unique fields. There was a bug report about it (#199). This patch fixes the problem by running validation after ``full_hydrate``.


SeanHayes:
`is_valid()` is now called during `save()`

Issue #11, opened 2010-09-14
Handling exceptions


rainerborene:
Hello, Daniel.

First I would like to say that tastypie is awesome. I'm developing an API for a web app, and I have a suggestion.

Suppose we have a function like this:

```
def get_detail(self, request, **kwargs):
    try:
        room = Room.objects.get(pk=kwargs.get('pk'))
    except Room.DoesNotExist:
        return HttpGone('Room entry does not exist.')
```

It would be great if tastypie treats **any** exception automatically.

Instead we can use the following code:

```
Room.objects.get(pk=param)
```

Many APIs out there have this functionality. Flickr, for example, has an error response.
[http://www.flickr.com/services/api/response.rest.html](http://www.flickr.com/services/api/response.rest.html)

I've extended the Resource class to solve that problem. See [EnhancedResource](http://gist.github.com/579511).

What do you think?

PR #274, opened 2011-10-25
Use exception message to describe Error 500 instead of 'This request could not be processed'.

claudiobi:
This patch expands on https://github.com/toastdriven/django-tastypie/commit/93ae0b3e for issue #11.

Whenever an exception raises an HTTP 500 Error, and that exception comes with a message, show *that message* as a description of the HTTP 500 Error, instead of a generic "Sorry, this request could not be processed. Please try again later."

### DIFF
--- a/tastypie/resources.py
+++ b/tastypie/resources.py
@@ -254,6 +254,11 @@ class Resource(object):
         data = {
             "error_message": getattr(settings, 'TASTYPIE_CANNED_ERROR', "Sorry, this request could not be processed. Please try again later."),
         }
+
+        # If the exception comes with an error, show that instead of "Sorry, this request..."
+        if getattr(exception, 'message', None):
+            data["error_message"] = exception.message
+
         desired_format = self.determine_format(request)
         serialized = self.serialize(request, data, desired_format)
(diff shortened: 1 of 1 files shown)


tastypie/resources.py
```diff
@@ -254,6 +254,11 @@ def _handle_500(self, request, exception):
         data = {
             "error_message": getattr(settings, 'TASTYPIE_CANNED_ERROR', "Sorry, this request could not be processed. Please try again later."),
         }
+
+        # If the exception comes with an error, show that instead of "Sorry, this request..."
+        if getattr(exception, 'message', None):
```
georgedorn: Maybe only do this for exceptions that also have the attribute "safe_for_public_viewing" or similar?  That way you can throw custom exceptions with that attribute set and expect to see them with DEBUG=False, but others are masked by default.

SeanHayes: :+1:

SeanHayes:
In the past parsing errors would result in 500 errors, but we now return the proper 4xx header. We also provide a hook function for customizing error responses when an exception is caught.

PR #299, opened 2011-11-22
Adding a LongField

paulcavallaro:
Hey TastyPie Devs,

First off, thanks a ton for creating such a useful django app. I ran into the need for a LongField personally, and though it would be a nice addition to tastypie. The code itself is pretty straightforward, most of it being tests that were cribbed off of alphanumeric.

Let me know if there's anything I should do differently.

Thanks,

-Paul

### DIFF
--- a/tastypie/fields.py
+++ b/tastypie/fields.py
@@ -222,6 +222,22 @@ class IntegerField(ApiField):
         return int(value)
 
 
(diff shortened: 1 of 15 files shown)


SeanHayes:
Note: int and long are now the same in Python 3, we may need to account for that.

PR #315, opened 2011-12-17
Add an authentication method based on a temporary token.

jorgeecardona:
This pull-request add a new authentication method based on a temporary token that it is valid only some amount of time. This method use the "WWW-Authenticate" and "Http-Authorization" headers. A new token must be created in some "start session endpoint", and in each session a 'Http-Authorization' header with: "Token xxx" must be send.

Valid time is set in settings.py with TOKEN_VALID_TIME

Tests and migrations are included.

### DIFF
--- a/docs/authentication_authorization.rst
+++ b/docs/authentication_authorization.rst
@@ -99,7 +99,7 @@ just for this purpose, so you'll need to sensure ``tastypie`` is in ``INSTALLED_
 To use this mechanism, the end user needs to specify an ``Authorization`` header. 
 Example::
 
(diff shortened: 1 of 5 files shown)


SeanHayes:
Using Django signatures or generating JWT might be better. I think developing those sorts of auth types might best be left as an exercise for the developer. Any of these auth types won't work out of the box the was ApiKey auth does, they require a more complicated flow (you have to first auth with username and password to get the token) that I'd rather not be responsible for. It might be better to revisit this if there's stronger demand and support for it.

https://docs.djangoproject.com/en/1.9/topics/signing/#using-the-low-level-api

PR #329, opened 2012-01-12
log exceptions regardless of settings.DEBUG

ross:
it's best if exceptions (via _handle_500) are logged regardless of settings.DEBUG. it's much easier to look at stack traces in the console or via file (if you have things configured to go to one) rather than always having to debug exceptions that have been returned to the calling client. 

it's not that big a deal when the calling client is a web browser (at least with jsonview,) but if you're calling with another device (iPad, android, ...) it's a pain to have to look there for the message.


SeanHayes:
Looks like mail_admins has been removed.

Related to #679.

SeanHayes:
Closed by #1376

PR #337, opened 2012-01-20
Propagate Django field attributes into ApiFields.

Unholster:
This patch makes it so that all undeclared resource field attributes are inherited from the underlying model field.

We build our API documentation from the schema view and the help_text attribute is really, well, helpful. It's a shame that when a resource field is declared and the help_text is not included in the declaration the existing help_text from the model field is not inherited.

This is useful to keep the schema view accurate without having to be overly redundant in the resource declarations.

Not sure the implementation is the cleanest, if it doesn't get accepted I hope it at least inspires a better version.

### DIFF
--- a/tastypie/resources.py
+++ b/tastypie/resources.py
@@ -1503,10 +1503,6 @@ class ModelResource(Resource):
             return final_fields
 
         for f in cls._meta.object_class._meta.fields:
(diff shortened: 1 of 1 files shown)


SeanHayes:
This has already been fixed.

PR #338, opened 2012-01-21
Added exception logging in Resource view wrapper.

ionelmc:


SeanHayes:
Related to #1320

tastypie/resources.py
```diff
@@ -200,10 +201,13 @@ def wrapper(request, *args, **kwargs):
 
                 return response
             except (BadRequest, fields.ApiFieldError), e:
+                logger.exception("BadRequest/ApiFieldError: ")
```
SeanHayes: These log lines don't actually log the exception though.

PR #354, opened 2012-02-06
except: in ModelResource.obj_update swallows (real) exceptions

pelme:
The "except:" statement in ModelResource.obj_update seems unnecessary. Exceptions from deep within Djangos ORM are swallowed (i.e. real errors), such as DatabaseError's.

It is good practice to use except: only in very rare situations (if any).

I cannot figure out what kind of exception would be expected here, and the tests does not give any clues since all tests pass without it. Therefore I assume that it is not required.

I discovered this problem when my manual tests against my API failed, but the very same testcase succeeded. I had missed to apply a migration, so Django raised a DatabaseError about an unknown column. That error was eaten in obj_update(), and later another unrelated exception was raised instead.


pelme:
Closing this, I do not intend to keep working on this PR.

PR #360, opened 2012-02-09
when in debug mode don't return before logging the error

ross:
500 errors aren't logged when DEBUG=True which can make it difficult to track down what's going on if the requests are coming from a device (iphone/android/etc) and can't easily see the details of the response content.


SeanHayes:
Fixed in eb9a5a82d7dcf3a1af65edb3822a3397f604912c

PR #378, opened 2012-02-17
Updating Serializer example in docs

nishadmusthafa:
Updated the CSV Serializer example which is now tested and works correctly.


SeanHayes:
This has already been fixed. Also, serializers shouldn't return responses.

PR #379, opened 2012-02-18
Modifying Error Handling of ModelResource.obj_get()

rdegges:
Currently, if `obj_get` raises a ValueError when trying to lookup the object,
tastypie will return a 404 not found error along with an error message saying
that the object type was wrong.

This behavior is undesirable, as the 404 status code doesn't accurately explain
what happened to make the request fail.

Instead, I propose raising a BadRequest exception (like the other ModelResource
methods do), which returns an HTTP 400 (BAD REQUEST) signaling to the user that
the request was malformed.

There are no tests attached to this change because I couldn't find any tests
that already exercise this part of the codebase. I'd be happy to add something
to this patch if necessary to get it accepted.

### DIFF
--- a/tastypie/resources.py
+++ b/tastypie/resources.py
@@ -1778,7 +1778,7 @@ class ModelResource(Resource):
 
             return object_list[0]
         except ValueError:
-            raise NotFound("Invalid resource lookup data provided (mismatched type).")
+            raise BadRequest("Invalid resource lookup data provided (mismatched type).")
 
     def obj_create(self, bundle, request=None, **kwargs):
         """


SeanHayes:
Hmm, I'm wondering if this should be merged or not. Technically, NotFound is more specific than BadRequest is, and given that this error is raised in obj_get, I think it might be possible to see it during GET requests on a bad URL and not just when sending bad data in the request body.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

> 400 Bad Request
>
> The request could not be understood by the server due to malformed syntax. The client SHOULD NOT repeat the request without modifications.

> 404 Not Found
> 
> The server has not found anything matching the Request-URI.

In the case of a bad GET request, I think 404 is more appropriate, though in the case of POSTing, PUTting, or PATCHing bad URLs to the server for field data, then 400 makes more sense.

PR #395, opened 2012-03-04
Allow negative primary keys

ghost:
Hi toastdriven (and others),

This commit allows primary keys to be negative (which Django _does_ allow). I've added a small unit-test and provided them with (hopefully) useful comments.

Kind regards,
Martijn


SeanHayes:
Our urls now accept any characters.

PR #429, opened 2012-03-31
Transaction middleware

### LINKED ISSUE #85: TransactionMiddleware not called when exceptions occur during Tastypie processing
Because Tastypie is intercepting exceptions to provide a proper client response, it looks like the TransactionMiddleware is not called properly.  This causes a partial commit of data rather than a rollback.

Steps to reproduce:
1. Ensure 'django.middleware.transaction.TransactionMiddleware' is in your MIDDLEWARE_CLASSES.
2. Samples:

model.py

```
(imports skipped)
class TestModel(models.Model):
    name = models.CharField(max_length=50, unique=True)
```

---

resource.py

```
(imports skipped)
class TestResource(ModelResource):
    class Meta:
        queryset = TestModel.objects.all()
        resource_name = 'test'
        allowed_methods = ['get','post']

    def obj_create(self, bundle, request=None, **kwargs):
        # Ignore bundle data for this simple test
        testA = TestModel(name='testA')
        testA.save()

        testB = TestModel(name='testA')
        testB.save()  # Should throw db exception and roll everything back

        bundle.obj = testA

        return bundle
```

---

urls.py

```
(imports skipped)
v1_api = Api(api_name='v1')
v1_api.register(TestResource())

urlpatterns = patterns('',
    (r'^api/', include(v1_api.urls)),
)
```
1. Fire up Django and POST to the API:
   
   ```
   curl -sv0 \
   -H 'Content-Type: application/json' \
   -X POST \
   -d "{\"name\":\"not used\"}" \
   http://localhost:8000/api/v1/test/ 
   ```
2. Request should error with a unique constraint violation, and the database should be empty.  Currently however, "testA" is created.

vbabiy:
The idea here is to use the the tastypie transactional middleware and it will handle both regular requests and tastypie 

If you guy like this I will complete the pull requests with docs and tests, this should fix #85.

Let me know what you guys think?


georgedorn:
The main problem with the tests is that we're running the test suite against a non-transactional database.  I'm going to open an issue for this because Travis supports other databases and we should really be testing against production database engine(s), not SQLite.

We should also have a test confirming the partial-save issue, since that is by far the largest concern.  There's some suggestion that the issue doesn't affect Postgres users, but it'd be good to be able to prove it (and prove the fix works for MySQL and other dbs).

SeanHayes:
TransactionMiddleware is now deprecated: https://docs.djangoproject.com/en/1.6/topics/db/transactions/#transaction-middleware

PR #433, opened 2012-04-06
avoid double query on the_m2ms, 715 & 726

ross:
the_m2ms is queried twice, once in an if and again in the for m2m in the_m2ms. if the_m2ms is None prevents the extra query, but may or may not be valid depending on what all values the_m2ms could hold.


SeanHayes:
Added, thanks!

PR #435, opened 2012-04-09
Do not return 404 when an uncaught ObjectDoesNotExist is raised

pelme:
Currently, any uncaught DoesNotExist that gets raised in user hydrate() methods is not reported as 500 errors.

Django reports those exceptions as 500 errors, and tastypie should do that too to avoid confusion. I had uncaught DoesNotExist exceptions in my custom hydrate methods and did not notice the error before users complained, since it never got logged as a 500 error.

Removing the special casing for ObjectDoesNotExist does not yield any regressions in the test suite, so I guess it was never really intended to be there.


SeanHayes:
I agree, tastypie already catches it's own DoesNotExist errors, we shouldn't catch rogue errors as well.

SeanHayes:
Should fix #194.

pelme:
Closing this PR, I do not intend to keep working on it.

Issue #452, opened 2012-04-18
alter_detail_data_to_serialize not called in get_multiple


danic:
I was overriding alter_detail_data_to_serialize() and noticed my resource subclass behavior was correct for a detail GET but not when getting a subset of a resource list, e.g. "http://localhost:8000/api/v1/entry/set/1;3/".

On commit 7d998a2b6002f0f11de8fdefbb2c9f46c8247c88 (latest as of the time of this writing), I fixed this with the following change in resources.py:

diff --git a/django-tastypie-0.9.11-7d998a2b60/tastypie/resources.py b/django-tastypie-0.9.11-7d998a2b60/tastypie/resources.py
--- a/django-tastypie-0.9.11-7d998a2b60/tastypie/resources.py
+++ b/django-tastypie-0.9.11-7d998a2b60/tastypie/resources.py
@@ -1394,6 +1394,7 @@
                 obj = self.obj_get(request, pk=pk)
                 bundle = self.build_bundle(obj=obj, request=request)
                 bundle = self.full_dehydrate(bundle)
-                bundle = self.alter_detail_data_to_serialize(request, bundle)
               objects.append(bundle)
           except ObjectDoesNotExist:
               not_found.append(pk)

I'm assuming alter_detail_data_to_serialize() should get called in get_multiple() the same way it is in say get_detail () ?

PR #453, opened 2012-04-19
Issue #452

danic:
I could be way off but I think a call to alter_detail_data_to_serialize  is missing from get_multiple in resources.py


SeanHayes:
Either that or alter_list_data_to_serialize. Arguments could be made for either I guess.

PR #454, opened 2012-04-20
Fix crash when a non-existent field is entered on a ForeignKey resource

pelme:
Currently it is possible to pass a non existent field in the data in a foreign key resource, which does not exist.

tastypie should ignore that data, but currently that field is tried to be looked up, which will fail with an AttributeError since it does not exist. See the test case for an example.


SeanHayes:
Fixed in #1336

PR #465, opened 2012-04-30
New cookbook example for using resources in views with a valid resource_uri.

yesimon:
The cookbook example for using resources in views will generate an object with no valid ``resource_uri`` because it is not associated with ``tastypie.Api`` instance. This docs update shows how you can get a valid ``resource_uri`` field on an resource in views.


SeanHayes:
This has been added to master.

PR #499, opened 2012-05-19
Added ability to return related resource primary key in addition to uri or "full".

justinzane:
Some clients fail to deal with resource URIs properly. [ExtJS4, I'm looking at you.] To allow these clients to deal with relationships, they require the primary key of the related model. This is simply a tweak to allow this to work for anyone.


SeanHayes:
This seems like non-standard behavior. I'd recommend fixing the client, stripping the pk from the resource_uri on the client, or using the workaround @pztrick suggested.

PR #532, opened 2012-06-15
Make sure TestApiClient doesn't send unicode (proposed fix for #531)

### LINKED ISSUE #531: TestApiClient.patch() sends unicode data which FakePayload doesn't handle
I didn't really have time to investigate this fully, but I think this is related to https://code.djangoproject.com/ticket/10571

Seems to work fine if I convert the output of the serializer to a string

ghing:
FakePayload doesn't seem to handle unicode

Fixes #531


SeanHayes:
Superceded by #1433

PR #574, opened 2012-07-17
[Feature] Give the Authentication classes an ability to whitelist some methods

dedsm:
use case: having a UserResource where a registration and a get_detail could be done by a guest user, and for every other method we need authentication.


SeanHayes:
Going to close this, this can already be done by overriding `is_authenticated` or by doing these checks in an Authorization class.

PR #595, opened 2012-08-02
Custom name for objects in a collection

ebertti:
Just another attribute inside of Meta class of Resource

### DIFF
--- a/tastypie/resources.py
+++ b/tastypie/resources.py
@@ -80,6 +80,7 @@ class ResourceOptions(object):
     include_absolute_url = False
     always_return_data = False
     collection_name = 'objects'
+    object_name = None
     detail_uri_name = 'pk'
 
     def __new__(cls, meta=None):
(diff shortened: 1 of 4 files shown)


tastypie/resources.py
```diff
@@ -80,6 +80,7 @@ class ResourceOptions(object):
     include_absolute_url = False
     always_return_data = False
     collection_name = 'objects'
+    object_name = None
```
SeanHayes: This is never set anywhere.

SeanHayes:
We'll need to know what the use case is.

ebertti:
Better customization and integrations with legace systems

SeanHayes:
Could you provide an example? What are you trying to do? You can already define different field names on your Resources than you have on your models. The case for adding a new attribute which we'll have to support, and which will only be used by a single serializer, doesn't seem strong enough to me.

SeanHayes:
Wait, I just saw the Gist, I get it now. I'll try to think if there's a way we can do this.

SeanHayes:
#1380

Issue #601, opened 2012-08-07
Extra fields in PUT requests are silently ignored


tolomea:
If the API client includes an extra invalid field in the put request (perhaps due to a typo) no error is raised. Additionally under "always_return_data = True" the invalid field is echo'd back to the client.

SeanHayes:
I'll probably modify this to make it optional.

SeanHayes:
Duplicate of #596

sshishov:
Why there is partial PUT here? Partial PUT should be PATCH in my opinion...
I provide incomplete data in PUT which should raise the exception but I got 200 results and fields which have not been provided remained untouched but they should be changed to default value if any or raise the error that the field is required and 400 response!

PR #624, opened 2012-08-14
Raising an error when a PUT or POST includes an unlisted field

maraujop:
Avoids silent bugs when a consumer submits data to a Tastypie API. This could be toggled by a Tastypie setting variable, though I like the idea of raising an error by default.

This breaks one test in core: 
https://github.com/toastdriven/django-tastypie/blob/master/tests/core/tests/resources.py#L1766

Related to GH-601 and GH-615


SeanHayes:
For this change to be accepted:
1. Collect all extra fields, raise error including all extra field names
2. Needs to be optional, controlled by a setting
3. Have a unit test demonstrating extra data is allowed when turned off, not allowed when turned on.

PR #634, opened 2012-08-21
Allow registering of Resources the DRY way

dekoza:
This patch allows for more DRY-compliant registering of Resources.

This is compatible with current way (registering of instances) but also allows for registering of classes only and classes/instances provided in an iterable.

Example usage:

```python
from tastypie.api import Api
# and other imports 

v1 = Api(api_name='v1')

# define Resources
# (...)

v1.register(FirstResource)
v1.register(SecondResource())
v1.register([ThirdResource(), FourthResource])
```


SeanHayes:
I'm not sure this is worth the complication to our tests.

It's pretty easy to create a resource already, and letting users create a resource merely by specifying a model class seems like it could lead to security issues.

PR #649, opened 2012-09-03
ValueError handled in deserialize method

piotrbulinski:
When I do POST with bad JSON, I receive 500. Wouldn't it be nicer to handle it in deserialize@resources.py, so I'd get for eg.:

400 BAD REQUEST:
Invalid data sent: Expecting , delimiter: line 1 column 180 (char 180)


SeanHayes:
Fixed in #1373.

Issue #591, opened 2012-08-01
Django DateField is assigned as DateTimeField


akoumjian:
https://github.com/toastdriven/django-tastypie/blob/master/tastypie/resources.py#L1553

When the ModelResource fields are created, this sets the DateField type to a tastypie DateTimeField. This causes it to be improperly handled in some instances and can break form validation when used with the model.

nutztherookie:
is there a solution to this problem?

SeanHayes:
Fixed in b248e7fda8c029bf0309b5e1f0e9db99328d5642

PR #664, opened 2012-09-08
Map Django's DateField to Tastypie's DateField

dirn:
When using a `DateField` in a `Resource` it can be a bit misleading when a datetime is used in the response. This can also lead to validation errors when linking to a `ModelResource`. This introduces the idea of mapping them to `DateField` instead of `DateTimeField`.

As a semantic change, `tastypie.utils.timezone.aware_date()` will now return a date object instead of a datetime.

Unfortunately these changes can potentially be backwards incompatible. This is noted in the log.

This is for #591.


SeanHayes:
Fixed in b248e7fda8c029bf0309b5e1f0e9db99328d5642

Issue #524, opened 2012-06-13
Allow exclude filtering


vesterbaek:
Sometimes it is usefull to be able to exclude instead of just filtering. I've added a simple solution, where prepending a field name with '!' in the url results in an exclude

For example, to filter(id=42)

```
/api/v1/model/?id=42
```

and to exclude(id=42)

```
/api/v1/model/?!id=42
```

I'm not sure it's the best syntax, but it's what I've come up with that's simple and easy to implement. I've overridden the following in my own resources to do the filtering

```
    def build_filters(self, filters=None):
        if not filters:
            return filters

        applicable_filters = {}
        # Normal filtering
        filter_params = dict([(x, filters[x]) for x in filter(lambda x: not x.startswith('!'), filters)])
        applicable_filters['filter'] = super(MyResource, self).build_filters(filter_params)
        # Exclude filtering
        exclude_params =  dict([(x[1:], filters[x]) for x in filter(lambda x: x.startswith('!'), filters)])
        applicable_filters['exclude'] = super(MyResource, self).build_filters(exclude_params)

        return applicable_filters

    def apply_filters(self, request, applicable_filters):
        objects = self.get_object_list(request)

        f = applicable_filters.get('filter')
        if f:
            objects = objects.filter(**f)
        e = applicable_filters.get('exclude')
        if e:
            objects = objects.exclude(**e)
        return objects
```

Input is welcome

doingandlearning:
This issue has been open for a while and is the same as issue #726. I suggest closing #726 and creating a PR for one of the solutions mentioned here. @imiric has a commit which seems to be the logical conclusion of the the discussion above - avoiding the '!' syntax and preferring '__ne' to allow for compatiability with other frameworks and to stay as close the ORM as possible. 

@imiric - do you want to create a PR for this or has this already been solved elsewhere now?

PR #726, opened 2012-11-16
Added a possibility to negate filters

KoFish:
By adding a '!' as the first character of a value in a filter it will instead be regarded as an `exclude`. For example `host:8000/api/user/?username=!bepa` will list all `resource` that doesn't have the username `bepa`.


doingandlearning:
This issue has been explored in more depth on #524. 
Can I suggest we close this thread and move the conversation to there to stop duplication?
Thanks.

PR #752, opened 2013-01-06
Update docs/resources.rst

orges:
bundle.request.user.pk == bundle.obj.pk condition check
leads to AttributeError while accessing by Anonymous user.
I think supressing AttributeError is mandatory
for all dehydrate methods.


SeanHayes:
AnonymousUser.pk is None: https://docs.djangoproject.com/en/1.9/ref/contrib/auth/#anonymoususer-object

Maybe request.user never got set.

PR #773, opened 2013-01-18
Remove unreachable branch in ApiKeyAuthentication

gsamokovarov:
The condition is always guaranteed to be `False` by the outer if statement. I have also run all of the tests for the Python's I have an they seem to be running fine.


SeanHayes:
This has already been fixed.

PR #778, opened 2013-01-23
Fixed CSVSerializer example's `to_csv` method to actually work.

eykd:
I tried using the example as provided in the current docs, and found that the `to_csv` method didn't work. I've fixed this, based on my own working version. I have to tested the `from_csv` method, but I did check it against the docs, and it appears correct.

The improved example also demonstrates better subclassing of parent Serializer, so the subclass doesn't need to know which formats are already implemented on the parent class.

### DIFF
--- a/docs/serialization.rst
+++ b/docs/serialization.rst
@@ -102,17 +102,36 @@ like::
 
 
     class CSVSerializer(Serializer):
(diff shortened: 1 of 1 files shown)


eykd:
@SeanHayes Thanks.

SeanHayes:
@eykd Thank you!

PR #785, opened 2013-01-26
Fix filter_value_to_python

MiguelCarranza:
Currently, it is not possible to use __in filters when a list is provided to Resource.get_list through kwargs.

The following code produces an error ("int() argument must be a string or a number, not 'list'")

```python
def get_concepts(self, request, **kwargs):
    concept_resource = ConceptResource()
    # ids actually generated in a different way 
    ids = ["1", "2", "3", "4"]
    return concept_resource.get_list(request, id__in=ids)
```

Basically, this pull request allows to use a list like 'ids' in kwargs, so that we could filter by, for instance, a set of primary keys.


georgedorn:
This is a useful feature but the PR went unaddressed for too long, so I had to update it to allow merging and running on django 1.6+.  See #1289.

Issue #793, opened 2013-02-01
Allow Meta.fields to be empty list to specify no field introspection


warvariuc:
> fields
Controls what introspected fields the Resource should include. A whitelist of fields. Default is []

Please mention that if Meta.fields == [] it means that ALL Django model fields are introspected. To simulate no introspection you can do `Meta.fields = ['']`
http://stackoverflow.com/questions/14597332/tastypie-custom-fields

SeanHayes:
I'll update the code to support `Meta.fields = None`, since it's cleaner.

SeanHayes:
Actually, tastypie should probably obey when a user explicitly provides an empty list.

PR #794, opened 2013-02-04
Clarified behavior of Meta.fields with empty list

marblar:
A documentation suggestion from #793


SeanHayes:
I changed the way `fields = []` works in #793 and updated these docs.

PR #814, opened 2013-02-14
Fix for related fields with "blank=False" option.

dbajal:
https://github.com/toastdriven/django-tastypie/issues/780


SeanHayes:
I think this was fixed in d959440bde59c692a0578671519d26a1b95fbe5f.

PR #815, opened 2013-02-14
More pervasive support for kwargs from urlpatterns.

kennknowles:
Hello!

This is actually bug report/feature request with a failing test case. I know how to fix it, essentially, but would like feedback.

The issue: I am working on a project (1) that has a URL structure like this:

```python
urlpatterns = patterns('',
   (r'^a/(?P<domain>[a-z]+)/api/', include(api.urls)),
)
```

Each resource depends on the `domain` parameter extensively, which is passed to the resource views as a kwarg (and could be added to the bundle, etc.)

However, these kwargs are not passed to `get_resource_uri` on [this line in `get_list`](https://github.com/toastdriven/django-tastypie/blob/master/tastypie/resources.py#L1262) and in fact `resource_uri_kwargs` is not written to incorporate them, so the [`reverse` lookup on this line](https://github.com/toastdriven/django-tastypie/blob/master/tastypie/resources.py#L304) always fails. This manifests as missing `next` and `prev`.

This cannot be fixed by adding it to the bundle because the bundle is not passed to `get_resource_uri` in the case of a list endpoint, and it cannot be in the current design. (2)

This can be fixed by passing the `kwargs` everywhere reliably and modifying `resource_uri_kwargs` to include them.

Note that as far as I can tell this _cannot_ be done by overrides except overriding `get_list` because that is where the problematic invocation of `get_resource_uri` lies. But `get_list` has a lot of functionality and is not an appropriate override point. I don't think this should require any overrides.

I am happy to go ahead and make the necessary changes; please comment.

Thanks,

Kenn

(1) The project is open source and available [on github](https://github.com/dimagi/commcare-hq) but very large, so I have presented a simplified example.

(2) This merging of two independent functions based on whether `bundle_or_obj` is `None` seems to needlessly complicate things. It seems that [in the past `get_resource_uri` and `get_resource_list_uri` were separate](https://github.com/toastdriven/django-tastypie/blob/v0.9.11/tastypie/resources.py#L601). This makes sense and would make it simple to pass the bundle to `get_resource_list_uri`.

### DIFF
--- a/tests/core/tests/api.py
+++ b/tests/core/tests/api.py
@@ -18,20 +18,6 @@ class UserResource(ModelResource):
         resource_name = 'users'
         queryset = User.objects.all()
 
-class ParameterizedNoteResource(ModelResource):
-    """
-    A resource that requires the kwarg `param` to be
-    passed by the URL pattern.
-    """
-    
-    #def obj_get_list(self, bundle, **kwargs):
-    #    param = kwargs['param']
-    #    return super(ParameterizedNoteResource, self).obj_get_list(self, bundle, **kwargs)
-
-    class Meta:
-        resource_name = 'notes'
-        queryset = Note.objects.all()
-
 
 class ApiTestCase(TestCase):
     urls = 'core.tests.api_urls'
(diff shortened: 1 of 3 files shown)


kennknowles:
(closing my stale PRs - feel free to make whatever use you like of this)

PR #856, opened 2013-03-10
Update cookbook.rst - Nested resources & bundle

ad-m:
According to http://stackoverflow.com/questions/15157071/tastypie-nested-resources-cached-obj-get-takes-exactly-2-arguments-1-given


SeanHayes:
Another fix has already been merged.

PR #862, opened 2013-03-13
In DecimalField.hydrate, catch decimal.InvalidOperation and raise ApiFieldError

kennknowles:
Current behavior when invalid data is submitted to `DecimalField.hydrate` is to let the `InvalidOperation` error escape to the top level, resulting in a 500 error. This patch adds tests to ensure that this does not happen and, analogously to invalid date formats, raises an `ApiFieldError`.


SeanHayes:
Thanks!

PR #892, opened 2013-04-05
Returned model's 'verbose_name' in resource.

RossiAndrea:
This patch introduce field's `verbose_name` into API resource.

I've found this particularly useful in conjunction with Backgrid, to generate a "pretty labeled" table directly from api.


SeanHayes:
Fixed by #1370

PR #918, opened 2013-04-25
Update models.py

Fandekasp:
loaddata breaks because of the tastypie signal. This is probably a bug in django, but until it gets fixed, the solution is just to disable signals on loaddata.


SeanHayes:
This has been fixed, thanks for reporting!

PR #923, opened 2013-04-29
Update to Allow CleanedDataFormValidation to work with forms/models that have related fields.

chander:
...ationships to work when using CleanedDataFormValidation with django ORM models that have Related fields

Formerly, when an insert or update operation occurs on a model that uses a FK relationship field in the form, TastyPie would pass into the form validation the URI that was provided in bundle.data.  Unfortunately, Django really wants a PK reference (the output of model_to_dict, in this case) and would fail in such cases.  This fix detects those related fields in the target model (bundle.obj) and leaves the references to the related table intact, but still applies bundle values for all other fields (some of which might not be in the model, but might be required for the form.)

### DIFF
--- a/tastypie/validation.py
+++ b/tastypie/validation.py
@@ -1,7 +1,7 @@
 from django.core.exceptions import ImproperlyConfigured
 from django.forms import ModelForm
 from django.forms.models import model_to_dict
-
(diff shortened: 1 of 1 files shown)


mrooney:
This would be awesome!

PR #944, opened 2013-05-23
Made resources.py:deserialize respect the format argument

grischa:
Hi
the deserialize function in resources.py did not respect the format argument.
I am not clear why this is so. I found the need to set it manually when handling multipart POSTs, so I changed it
I ran the tests and all passed OK.


SeanHayes:
Fixed in 3a67470eb3575ba5afdddf64b4233e597e36f648, duplicated by 1354

PR #951, opened 2013-06-03
Check if data is valid before saving the object

unho:
The `is_valid` method in `Validation` classes (https://github.com/toastdriven/django-tastypie/blob/master/tastypie/validation.py) only return:

* An empty dictionary, if everything went fine,
* Or a dictionary with errors, if validation failed. Note that no exception is raised.

This methods are called in the `is_valid` method in `Resource` class (https://github.com/toastdriven/django-tastypie/blob/master/tastypie/resources.py#L1254). This method returns:

* True, if everything went fine,
* False, if any error were detected. Note that no exception is raised.

Finally this last method is called in `save` method in `Resource` classes. But there is no check to see if the data is valid and abort the saving: https://github.com/toastdriven/django-tastypie/blob/master/tastypie/resources.py#L2281


SeanHayes:
Resource.is_valid() sets bundle.errors, which we check for in the ModelResource.save() method.

PR #952, opened 2013-06-05
Cookbook: Added better example for using non-PK data

duebbert:
The original example doesn't work well in my opinion. The URIs are missing and the old URL (i.e. with the pk) still works which is surely in most cases not desirable. I've done my example so that theoretically it could be included in the main code base because it's backwards compatible if Meta.pk_field is not set. But I haven't done tests etc, so think it might be better in the cookbook for now to help other new users.


SeanHayes:
I think `detail_uri_name = 'profile__uuid'` is all you'd need.

PR #987, opened 2013-07-21
Update models.py

whardier:
Not sure when we would ever hit a 256 limit if we are using sha.  This 256 breaks MySQLs brain as well when adding unicode aware indexes.  255 doesn't.


SeanHayes:
Looks like the duplicate PR was merged 2 years ago, without a migration.

PR #997, opened 2013-08-05
Fix south migrations error

bendemboski:
Importing unicode_literals causes a problem in South when importing from django.db.models.fields because then the name of the model class that is passed into __import__ is unicode, rather than a byte string. I'm not quite sure why this causes a problem importing from this module but not others. However, removing the unicode_literals import from these migrations fixes the issue, and it doesn't look like the migrations have any literals that need to be unicode.

### DIFF
--- a/tastypie/migrations/0001_initial.py
+++ b/tastypie/migrations/0001_initial.py
@@ -1,5 +1,4 @@
 # encoding: utf-8
-from __future__ import unicode_literals
 import datetime
 from south.db import db
 from south.v2 import SchemaMigration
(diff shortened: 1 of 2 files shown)


SeanHayes:
This can probably be merged once the conflicts are resolved.

SeanHayes:
The needed changes have been merged into master. Thanks!

benjaoming:
For the record, if anyone is wondering, this was never merged nor released and thus tastypie 0.12.2, the final version with South support cannot migrate forwards.

Hopefully we've all moved beyond South :)

PR #999, opened 2013-08-06
Raise ApiFieldError if null or blank values aren't set where required

kdazzle:
As it is, Tastypie sets fields that aren't allowed to be null as null, creating database integrity errors down the line. This patch raises an error, as seems to be the original intention. The code was also gussied up a bit.


SeanHayes:
Fixed in #1208

Issue #1009, opened 2013-08-18
Custom User field not getting read from compat.py during ApiAuthentication


devangmundhra:
I have a Django project with a custom user field. When using ApiAuthentication for a modelresource, at
https://github.com/toastdriven/django-tastypie/blob/master/tastypie/authentication.py#L183

the User field is None causing the User.DoesNotExist exception. However, if instead of doing `from tastypie.compat import User` I actually put in the code from compat.py here, the custom user model and username_field are set fine.

SeanHayes:
I think this has been fixed by another change.

PR #1010, opened 2013-08-18
Issue #1009 Custom User field not getting read from compat.py during ApiAuthentication

devangmundhra:
If using a custom User model with create_api_key signal, it is possible that 
        this module was not loaded properly (because the custom user model was not
        setup by then, see compat.py :ImproperlyConfigured exception for more details).
        In that case, reloading the module will get the correct custom User module
        (otherwise sometimes the User model is None)


SeanHayes:
Fixed

PR #1012, opened 2013-08-19
Add more User model Meta information to migrations

stefanw:
This gives more meta information like e.g. `db_table` to the the Meta field of the `AUTH_USER_MODEL` in the frozen migration models. [The frozen Meta attributes are explained here](http://south.readthedocs.org/en/latest/ormfreezing.html#frozen-meta-attributes).

I had problems migrating because the Meta information in the frozen models did not have the `db_table` attribute. South then tried to setup an incorrect ForeignKey and ended up with `django.db.utils.ProgrammingError: relation "account_user" does not exist`.

I'm not sure this is the right approach, though. One could argue that the User model can't be properly frozen in tastypie anyways, as it might be completely different. However, maybe this works alright for a common case.

(Also: I don't think `get_model_meta` of South is a public API.)

If this is not worth including, may it serve as a guide for others to fix their migration.

### DIFF
--- a/tastypie/migrations/0001_initial.py
+++ b/tastypie/migrations/0001_initial.py
@@ -3,8 +3,13 @@ from __future__ import unicode_literals
 import datetime
 from south.db import db
 from south.v2 import SchemaMigration
+from south.modelsinspector import get_model_meta
 from django.db import models
-from tastypie.compat import AUTH_USER_MODEL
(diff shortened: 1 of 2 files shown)


SeanHayes:
Since there's a workaround and we're soon dropping support for Django <1.7, I'm going to close this. Also, it looks like a duplicate PR for this may have been merged.

PR #1017, opened 2013-08-23
Fixing import error for django versions older than 1.5

david-thorman:


SeanHayes:
Django 1.5 is now the minimum supported version.

PR #1054, opened 2013-10-18
Suppress Content-Type header on HTTP 204 (see #111)

acdha:
This is a tidied up version of the patch in #111 with tests


denisvlr:
I ran into a bug with this:
The method `assertHttpAccepted` (which applies for both 202 HttpAccepted and 204 HttpNoContent) checks for the absence of `Content-Type`, however it is deleted only in HttpNoContent.

PR #1067, opened 2013-11-04
optimize to_simple for speed

bculberson:
After doing some performance analysis of tastypie I found some really quick wins in the to_simple function of the serializers.  This dramatically sped up our endpoints and I believe would be universally useful and applicable.

The more used datatypes are positioned at the top of the function to reduce usage of get_attr.  usage of has_attr was removed and multiple calls to get_attr also removed.  type was also used in place of get_attr where possible.


SeanHayes:
Thanks for the idea. This has been included in #1337.

PR #1069, opened 2013-11-05
formated string as unicode

jsanchezpando:
some model instance casted to string can contains unicode values.


SeanHayes:
The merge conflicts need to be fixed.

PR #1103, opened 2013-12-27
Fixed DecimalField to be compatible with py2.6

Chern:
float is not permitted as argument type in python 2.6


SeanHayes:
Thanks!

PR #1123, opened 2014-01-27
Elaborate on Resource.full_dehydrate doc

leo-the-manic:
Spent a good bit of time trying to find out where I could exclude a particular field in the detail view, but only for a POST request, on my model resource. Turned out that `full_dehydrate` had the info I needed but I had no idea what the `for_list` argument was supposed to be so I figured I'd flesh it out a little for the next guy.


SeanHayes:
Thanks!

PR #1127, opened 2014-02-05
Fix to save something but always update the related fields.

viniciuscainelli:
Verifying that the data passed to the related field are only fields to find and not to update.


georgedorn:
Re-opening PR to re-run tests.

PR #1141, opened 2014-03-11
Add 'primary_key' Field To Schema

systemsoverload:
Re-created #1082 to reflect new branch/commits. 

This commit adds the field "primary_key" to the resource schema, it can be very useful for auto-generated UI/clients when using explicit primary key columns on your django models.


SeanHayes:
Sorry this took so long, thanks!

PR #1143, opened 2014-03-20
Removing documentation references to non-existent method is_authorized

jcwilson:
As far as I can tell, the is_authorized method does not exist and is never called


SeanHayes:
Thanks!

PR #1146, opened 2014-03-24
Django 1.6 tests

georgedorn:
Warning: this doesn't fix broken tests, it merely exposes them.  We haven't been running the majority of tests under django 1.6 for quite some time, possibly forever.  This is because 1.6's DiscoverRunner doesn't discover tests in modules other than tests.py or tests.test_*.py.


georgedorn:
Closing due to age; the DiscoverRunner appears to find tests when explicitly called via run_all_tests.sh, so this isn't needed.

Issue #1150, opened 2014-04-01
Concurrent POST / PATCH Race Condition in Resource.save() & save_m2m()


quietlyconfident:
Possibly similar to issue #742 .

I'm seeing errors due to a race condition between closely-spaced-in-time POST / PATCH requests involving a new child and existing parent object.

Situation: 
Child  w/ FK to Parent
Parent w/ reverse name to child

In my client-side app, I create a new Child object and make an unrelated modification to an existing Parent object (updating timestamp).

To persist this data to the server, I POST the Child to its endpoint and PATCH the Parent to its endpoint.  Thus, from the point of view of the user, the expected behavior is that:
- The child is created, and has a FK to the parent
- The parent's unrelated field is updated
- At the end of these two operations, in whichever order they the child should exist and have its FK, the parent should have its other field updated, and the parent should see the child on its reverse name relationship.

However, what actually seems to happen, rarely if the objects are about the same size, always if the Parent is much larger than the child (except NEVER when the --nothreading option is used on the Django test server) is that:
- The requests hit the server at approximately the same time.
- The Child object is successfully created and the server responds
- Then Parent object is successfully updated and the server responds; but in the course of this update, the Child object's FK to the parent object is deleted.

Investigating this with the debugger, it seems that the Resource.save() function saves bundles resulting from PATCH requests by retrieving all the object data, overlaying the data from the PATCH, and saving the whole thing.  So it seems like what is happening is:
- The requests hit the server
- The PATCH request begins processing and retrieves its object.
- The POST request finishes processing and the server responds
- The PATCH request overlays its data (a subset of the fields) onto the retrieved object data (which is, at this point, stale) and saves.  Because the reverse name field for the Child relationship is now different (i.e. it is the pre-POST data) it changes the Parent object BACK to how it was before the child's post.

While there are definitely client-side ways to work-around this issue (e.g. waiting for one request to finish before sending the next one; sending the reverse-name field in the PATCH), these are both rather wasteful in terms of time / server resources.

bates-hu:
face the same issue. The patch using save without specific fields is insane. when a update is finish between a patch, the result of the update will be overwrite.

Poshi:
2024 and facing the same issue. I'm trying to PATCH an object with inputs from different processes. Each process issues its own PATCH. Many of them arrive very close in time, almost simultaneously. The net result is that all PATCH are applied, but all but the first overwrite the value of the previous PATCH with the original value.
[...]

PR #1151, opened 2014-04-02
Fix race condition between POST / PATCH resources by using update_fields...

quietlyconfident:
Fixes issue #1150  by:
- Creating an update_fields list on the bundle during a patch
- Creating an m2m_update_fields list on the bundle during a patch
- During save, passing arguments to save to only update the changed data: save(update_fields=bundle.update_fields) 
- During save_m2m , don't save fields unless they appear in m2m_fields


SeanHayes:
https://travis-ci.org/django-tastypie/django-tastypie/builds/41000585

EndermanAPM:
I've _stole_ this idea to create a custom resource that overrides the methods and it works, would be great to add this to the tastypie codebase.

After reading the source for the whole afternoon to find a solution, I don't understand why you shoehorn the patch method into the put instead of doing the reverse, and only updating the provided fields. Wouldn't be easier to convert the PUT in a "full PATCH"?

Poshi:
I've been hitting the issue related to this PR during the last days. And it looks like there's already a solution offered! Could someone take a look at this and merge the PR for the next release?
That would solve a quite relevant bug, as the code in its current state breaks the compliance with their corresponding [RFC](https://datatracker.ietf.org/doc/html/rfc5789#section-2):

> The server MUST apply the entire set of changes atomically and never provide (e.g., in response to a GET during this operation) a partially modified representation.

georgedorn:
This PR is ten years old and cannot possibly be merged.  It couldn't be merged when it was created as it didn't match the style (semicolons, no spaces between args, other flake8 issues), failed existing tests and didn't provide test coverage for the new behavior.  The original author didn't come back to rectify any of that in the last decade.

The best I can do at the moment is cherry-pick this PR into a new branch off `master` and at least address the style issues.  I have done so in #1671.  If somebody would like to take a crack at testing it and proving it works (ideally in the form of a test case that fails on `master` but passes on `quietlyconfident-master`) I am happy to review and merge it.

As another option, if @EndermanAPM  would like to contribute his rework as a PR, I'm happy to look at it, and @Poshi could try it out in his scenario to see if it solves his issue as well.

PR #1155, opened 2014-04-08
Remove override of bundle data with kwargs

pyepye:
This adds pk to PUT response even when it is excluded

### DIFF
--- a/tastypie/resources.py
+++ b/tastypie/resources.py
@@ -2091,8 +2091,6 @@ class BaseModelResource(Resource):
         """
         lookup_kwargs = {}
         bundle.obj = self.get_object_list(bundle.request).model()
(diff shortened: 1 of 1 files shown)


tastypie/resources.py
```diff
@@ -2091,8 +2091,6 @@ def lookup_kwargs_with_identifiers(self, bundle, kwargs):
         """
         lookup_kwargs = {}
         bundle.obj = self.get_object_list(bundle.request).model()
-        # Override data values, we rely on uri identifiers
-        bundle.data.update(kwargs)
```
SeanHayes: We can't remove this line. If 'pk' is showing up in your data we'll need to remove it elsewhere.

PR #1160, opened 2014-04-15
Dynamic auth user model pk name in initial migration.

aivins:
The **id** column can be something else if **AbstractUser** is in use. In our case it was **person_ptr_id**. This fixes it for me. Provided the same fix to django-reversion yesterday: https://github.com/etianen/django-reversion/pull/301

### DIFF
--- a/tastypie/migrations/0001_initial.py
+++ b/tastypie/migrations/0001_initial.py
@@ -6,6 +6,12 @@ from south.v2 import SchemaMigration
 from django.db import models
 from tastypie.compat import AUTH_USER_MODEL
 
+try:
+    from django.contrib.auth import get_user_model
+except ImportError: # django < 1.5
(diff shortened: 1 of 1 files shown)


tastypie/migrations/0001_initial.py
```diff
@@ -6,6 +6,12 @@
 from django.db import models
 from tastypie.compat import AUTH_USER_MODEL
 
+try:
+    from django.contrib.auth import get_user_model
+except ImportError: # django < 1.5
+    from django.contrib.auth.models import User
+else:
+    User = get_user_model()
```
SeanHayes: Could you use tastypie.compat.get_user_model() instead?

aivins: I'm sure you can, as long as you've got a reference to User.

PR #1163, opened 2014-04-18
Fix issue #1009

mpyatishev:


SeanHayes:
I think #1009 has been fixed by another change.

PR #1166, opened 2014-04-21
Do not send 404 exceptions via `got_request_exception`

teferi:
django-tastypie informs other apps of exceptions that occured in _handle_500, but 404 exceptions aren't
really errors and tastypie encourages to raise 404 for certain cases. 
It seems logical to send `got_request_exception` only under same conditions when error logging is done.

This pull request is motivated by sentry integration, as sentry relies on catching `got_request_exception` signals and get's lots of false positives from tastypie if a resource raises NotFound or Http404.

Is this suggestion reasonable?
Should I add tests to this PullRequest?


SeanHayes:
We should only call this for 5xx errors, but not any 4xx errors.

PR #1167, opened 2014-04-21
Update index.rst

crazy449:


SeanHayes:
I think it's correct the way it is.

PR #1180, opened 2014-06-12
Handle skipped m2m fields in patch method

rkojedzinszky:
Patch method is for partially updating a resource. This patch allows to update a resource skipping m2m fields also.


rkojedzinszky:
Actually I could not reproduce the issue I found nearly two years ago.

SeanHayes:
Thanks for the update!

PR #1185, opened 2014-07-07
api_field_from_django_field: use DateField for django DateField

rkojedzinszky:
This fixes a case, when modifying a resource with a DateField and with always_return_data=True, the returned field would have a time part in it.

### DIFF
--- a/tastypie/resources.py
+++ b/tastypie/resources.py
@@ -1790,7 +1790,9 @@ class BaseModelResource(Resource):
         result = default
         internal_type = f.get_internal_type()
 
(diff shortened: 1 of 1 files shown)


SeanHayes:
Tests need to be updated to reflect the new date formatting.

@georgedorn Do you know if there was a reason DateField wasn't being used in the first place?

georgedorn:
No clue.  I can only speculate that it may have had something to do with timezone awareness or the lack thereof around the time it was written.

SeanHayes:
Fixed by #1365

PR #1186, opened 2014-07-09
Fix 500 error when trying to create an object without supplying related field

jakecr:
Given a resource with a ToOneField having null=False, blank=False and no default. A POST request to this resource with the field missing should result in "The '%s' field has no data and doesn't allow a default or null value." error just like any other non-blank, non-null field. However, because django raises ObjectDoesNotExist when trying to access this attribute a 500 error is returned.


SeanHayes:
Might be related to #1208

SeanHayes:
This has been fixed.

PR #1192, opened 2014-07-23
Provide mixins for easier subclassing and combining

mitar:
Otherwise aProvide mixins for easier and cleaner subclassing and combining.


SeanHayes:
We'll need to remerge this branch and see how the tests turn out.

SeanHayes:
@georgedorn I fine with this as long as you are.

mitar:
I updated the branch.

tastypie/contrib/gis/resources.py
```diff
@@ -75,3 +76,9 @@ def filter_value_to_python(self, value, field_name, filters, filter_expr,
             except ValueError:
                 pass
         return value
+
+
+class ModelResource(GeometryModelMixin, resources.ModelResource):
```
georgedorn: I've generally found that when you subclass this further with addition mixins, having the Mixin first can create MRO issues.  I don't have a concrete example of this, however.

PR #1193, opened 2014-07-26
Switch order of check for callable and string

mitar:
This is necessary because in `apply_sorting` attribute can be only a string, so by first checking for callable and then string in dehydrate, one can define a callable string, which can then be used both in dehydrate and `apply_sorting`.

### DIFF
--- a/tastypie/fields.py
+++ b/tastypie/fields.py
@@ -718,7 +718,11 @@ class ToOneField(RelatedField):
         foreign_obj = None
         error_to_raise = None
 
-        if isinstance(self.attribute, six.string_types):
+        if callable(self.attribute):
+            previous_obj = bundle.obj
+            foreign_obj = self.attribute(bundle)
+
+        elif isinstance(self.attribute, six.string_types):
             attrs = self.attribute.split('__')
             foreign_obj = bundle.obj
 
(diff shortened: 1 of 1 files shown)


SeanHayes:
https://travis-ci.org/django-tastypie/django-tastypie/builds/30940322

SeanHayes:
I think we need an example use case.

mitar:
OK, this is pretty tricky. But we use this for such an use case:

```python
# We need a function which is also a string. Function is used in dehydrating
# ManyToManyField, string is used in apply_sorting to create an order_by argument.
class RelatedNodes(str):
    def __call__(self, bundle):
        # Add name to related nodes
        return bundle.obj.related_nodes.regpoint('config').registry_fields(
            name='core.general#name',
        )
```

And then in the resource itself, we can define such field:

```python
    related_nodes = api_fields.ManyToManyField(to=NodeResource, attribute=RelatedNodes('related_nodes__config_core_generalconfig__name'), full=True)
```

So the code in `RelatedNodes` is our dynamic schema ORM based on Django. So in this way we can by setting `attribute` allow a constant string and also a dynamic string.

mitar:
Tests seems to fail because of some other error, not one by this pull request?

SeanHayes:
@mitar Those are really old tests. Could you fix your PR so it can be merged? When you push the new tests should run and we'll know if it can be merged.

I guess as long as the existing tests pass, changing the order of the if statements doesn't matter that much, unless we advertise this functionality.

mitar:
Do you prefer rebase or merge of `master` here?

SeanHayes:
rebase if possible, thanks!

mitar:
Updated.

mitar:
Thanks.

PR #1203, opened 2014-08-21
Data leakage fix

SeanHayes:
I found some code that was storing data directly on Field objects, which get reused between requests. This ablity was undocumented and untested, and could lead to unpredicatable behavior if someone tried to use it.

### DIFF
--- a/tastypie/fields.py
+++ b/tastypie/fields.py
@@ -73,7 +73,6 @@ class ApiField(object):
         self.null = null
         self.blank = blank
         self.readonly = readonly
-        self.value = None # REMOVEME
         self.unique = unique
         self.use_in = 'all'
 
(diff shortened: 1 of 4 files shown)


tastypie/fields.py
```diff
@@ -491,7 +491,6 @@ def __init__(self, to, attribute, related_name=None, default=NOT_PROVIDED, null=
 
         if self.to == 'self':
             self.self_referential = True
-            self._to_class = self.__class__
```
SeanHayes: This is set in contribute_to_class()

tastypie/fields.py
```diff
@@ -727,9 +729,9 @@ def dehydrate(self, bundle, for_list=True):
 
             return None
 
-        self.fk_resource = self.get_related_resource(foreign_obj)
```
SeanHayes: self.fk_resource isn't used outside this function

tastypie/serializers.py
```diff
@@ -240,19 +240,6 @@ def to_simple(self, data, options):
             return dict((key, self.to_simple(val, options)) for (key, val) in data.items())
         elif isinstance(data, Bundle):
             return dict((key, self.to_simple(val, options)) for (key, val) in data.data.items())
-        elif hasattr(data, 'dehydrated_type'):
```
SeanHayes: This type doesn't appear to be used anywhere.

SeanHayes:
@georgedorn This one was opened a while ago but I thought someone else should take a look at it.

Issue #981, opened 2013-07-07
djang-tastypie-speficied-fields


dan-klasson:
I created a simple extension for Tastypie that only fetches and returns the fields you specify:

https://github.com/dan-klasson/django-tastypie-specified-fields

Posting it here in case someone finds it useful.

PR #1205, opened 2014-08-21
Client specified fields

SeanHayes:


SeanHayes:
Should switch to comma separated list instead of repeated name/value pairs.

tastypie/resources.py
```diff
@@ -834,9 +834,31 @@ def full_dehydrate(self, bundle, for_list=False):
...
         use_in = ['all', 'list' if for_list else 'detail']
+        
+        all_fields = set(self.fields.keys())
+
+        allowed_fields = None
+        excluded_fields = None
+        
+        if not bundle.related_obj:
+            if 'fields' in bundle.request.GET:
+                allowed_fields = bundle.request.GET.getlist('fields')
```
SeanHayes: instead of requiring `fields` to be respecified over and over, just use comma separation.

SeanHayes:
Related: #1132, #981

PR #1208, opened 2014-08-21
ApiFieldError on empty non-null field

SeanHayes:


SeanHayes:
Might be a duplicate of #999

PR #1213, opened 2014-08-29
Allow use of timeout=0 in Simple cache

### LINKED ISSUE #1212: SimpleCache: timeout=0 is ignored
I used to configure my Resource classes with `SimpleCache` with `timeout=0` to prevent django's cache from storing my views. Starting from version 0.11.0 this does not work anymore, `SimpleCache` is always configured with the default cache value.

pveglia:
Fixes #1212

### DIFF
--- a/tastypie/cache.py
+++ b/tastypie/cache.py
@@ -60,7 +60,7 @@ class SimpleCache(NoCache):
         """
         super(SimpleCache, self).__init__(*args, **kwargs)
         self.cache = get_cache(cache_name)
-        self.timeout = timeout or self.cache.default_timeout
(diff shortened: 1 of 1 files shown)


tastypie/cache.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, cache_name='default', timeout=None, public=None,
         """
         super(SimpleCache, self).__init__(*args, **kwargs)
         self.cache = get_cache(cache_name)
-        self.timeout = timeout or self.cache.default_timeout
+        self.timeout = timeout if timeout != None else self.cache.default_timeout
```
SeanHayes: Should be 'if timeout is not None else'.

PR #1233, opened 2014-09-22
Security fix to prevent jsonp by default

hmiguellima:
The callback parameter in the request query string should not activate the jsonp format without checking the serializer.formats.


georgedorn:
Closed via #1476

PR #1237, opened 2014-10-14
Fill 'accessed' field without overloading .save()

prosto-chewey:


SeanHayes:
The old code sets "accessed" every time the model is saved, but this new code will only set "accessed" when the model is created.

prosto-chewey:
Sorry, I was inattentive. May be create signal on `post_save`?

SeanHayes:
We'd need a pre_save signal. I don't think this really needs fixing though.

PR #1240, opened 2014-10-20
Fix #1239.

### LINKED ISSUE #1239: get_via_uri not working for alphabetic ids that contain the resource name
Suppose I have a user whose id is `'myusers'` calling `get_via_uri` with `/v1/users/myusers/` won't work as espected. 

The function will try to get the id by removing `/v1/users` from the string but since d6cf75b134ccad4a174cfebdf33cddc11183d7af it will start searching from the right and in the above case will find the wrong instance of `"users"` and remove the whole string.

pmelet:
rfind up to the beginning of the resource key, to allow prefix/resource_name/resource_name/ in case resource_name is a valid identifier for an object


SeanHayes:
https://travis-ci.org/django-tastypie/django-tastypie/builds/38473629

Issue #1035, opened 2013-09-23
save_m2m failes to save the same object to multiple related resources


gjcourt:
Because of the way that `bundle.objects_saved` works, object identifiers are not namespaced by their field names. This means that if I had a resource with multiple related fields of the same type, I could save an object that would persist in one related field and not the other.

Suppose you had an object

``` python
class UserResource(ModelResource):
    pass

class ForumResource(ModelResource):
    class Meta:
        always_return_data = True
    moderators = fields.ManyToMany(UserResource, 'moderators', 'forum_mods', full=True)
    members = fields.ManyToMany(UserResource, 'members', 'forum_members', full=True)
```

Now suppose that I issued a `POST` to create a new forum with the following data

``` python
class ApiTest(ResourceTestCase):
    def test_create_forum(self):
        response = self.api_client.post('/api/1/forum/', format='json', data={
            'name': 'Test Forum',
            'members': {'username': 'gjcourt'},
            'moderators': {'username': 'gjcourt'},
        })
        import json
        data = json.loads(response.content)
        self.assertEqual(len(data['members']), 1)
        self.assertEqual(len(data['moderators']), 1)
```

The following test would fail because `def save_m2m()` would have cached the first related save of the `User` gjcourt and skip saving it to the second related resource.
![11x](https://f.cloud.github.com/assets/356410/1193583/fd2005fe-2475-11e3-9806-e654ea9d2f5e.png)

PR #1244, opened 2014-10-30
PUTing a non-existant object in multiple m2m resource causes an error

JannKleen:
This is actually the test that was lined out in #1035 (cc: @gjcourt @SeanHayes )
The expected behaviour when sending a PUT request like the one below is that a new user resource gets created (like it does if you were to add the user to just one m2m resource):

```
        user_data = {
            'username': 'valid but unique',
            'email': 'valid.unique@exmaple.com',
            'password': 'junk',
            }
        forum_data = {'members': [user_data_1, ],
                      'moderators': [user_data_1, ]}

        request.path = reverse('api_dispatch_detail', kwargs={'pk': forum.pk,
                                                              'resource_name': resource._meta.resource_name,
                                                              'api_name': resource._meta.api_name})

        response = resource.put_detail(request)

```

### DIFF
--- a/tests/related_resource/tests.py
+++ b/tests/related_resource/tests.py
@@ -15,7 +15,7 @@ from core.tests.resources import HttpRequest
 
 from related_resource.api.resources import CategoryResource, ForumResource, FreshNoteResource, JobResource, NoteResource, PersonResource, UserResource
 from related_resource.api.urls import api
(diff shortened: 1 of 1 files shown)


xiaolong3105:
您好，您的邮件我已收到，我会尽快处理。
 
Best Wishes!
晓龙

PR #1245, opened 2014-10-30
prefetch_related causes api to return outdated data

JannKleen:
When issuing a PUT request on a prefetched field, the old data is returned.

A failing test is included. I'll try to fix this myself, but any tips are appreciated :smile: 

**Resource:**
```
    class Meta:
       always_return_data = True
       queryset = Forum.objects.prefetch_related('moderators', 'members')
```

**Pseudo-Code:**
```
    Forum.objects.create()
    request.put('/api/forum/1/', {'moderators': [{'username': 'JannKleen', }, ], })
```

This should return the new moderator, but instead returns an empty list (the old value of the m2m field).

### DIFF
--- a/tastypie/resources.py
+++ b/tastypie/resources.py
@@ -1430,6 +1430,9 @@ class Resource(six.with_metaclass(DeclarativeMetaclass)):
             if not self._meta.always_return_data:
                 return http.HttpNoContent()
             else:
+                # Invalidate prefetched_objects_cache for bundled object
+                # because we might have changed a prefetched field
+                updated_bundle.obj._prefetched_objects_cache = {}
                 updated_bundle = self.full_dehydrate(updated_bundle)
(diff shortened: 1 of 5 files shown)


tastypie/resources.py
```diff
@@ -1430,6 +1430,9 @@ def put_detail(self, request, **kwargs):
             if not self._meta.always_return_data:
                 return http.HttpNoContent()
             else:
+                # Invalidate prefetched_objects_cache for bundled object
+                # because we might have changed a prefetched field
+                updated_bundle.obj._prefetched_objects_cache = {}
```
JannKleen: Sorry for the delay ... when adding prefetch_related to the queryset (resources.py:213) to reduce the number of db requests, tastypie will return the prefetched results for M2M fields, even though we updated them in the request.

JannKleen: @georgedorn ping :wink:

PR #1256, opened 2014-12-02
Force attribute name to be string in Exception

rkojedzinszky:
When an exception is to be thrown, and previous_objects's representation containts unicode characters, as Django encodes them to utf-8, the string formatting dies with an unicode error. This fixes that.


SeanHayes:
I think this might fail in Python 3, since the str type is actually a unicode string but b'' strings are bytes strings.

Maybe instead you should make it a u'' string. We'll need a unittest to confirm the error though.

PR #1262, opened 2014-12-21
Improved admin UI for API Keys

gorillamania:
Improved admin UI for API Keys

* Make api keys searchable, by key or username
* Displays more information on the initial screen, with sortable columns
* Use raw_id_fields for user, because detail page can get very slow when there are a lot of api keys

api_key_admin
raw id field

### DIFF
--- a/tastypie/admin.py
+++ b/tastypie/admin.py
@@ -6,9 +6,10 @@ from django.contrib import admin
 if 'django.contrib.auth' in settings.INSTALLED_APPS:
     from tastypie.models import ApiKey
 
-    class ApiKeyInline(admin.StackedInline):
-        model = ApiKey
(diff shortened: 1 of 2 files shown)


SeanHayes:
This needs to be updated since the admin.py file has been changed.

tastypie/admin.py
```diff
@@ -6,9 +6,10 @@
 if 'django.contrib.auth' in settings.INSTALLED_APPS:
     from tastypie.models import ApiKey
 
-    class ApiKeyInline(admin.StackedInline):
```
SeanHayes: Need to keep this though

SeanHayes:
Fixed and merged, thanks!

PR #1265, opened 2014-12-28
Provide use_in value to the use_in field callback

mitar:


SeanHayes:
https://travis-ci.org/django-tastypie/django-tastypie/builds/45277029

Tests failed, backwards incompatible.

mitar:
Updated.

mitar:
Hm, why you close and reopen?

SeanHayes:
My mistake, I thought it was running against the old tests. The build hadn't started yet anyway.

SeanHayes:
@georgedorn  @mitar There's still the issue of this breaking for anyone who uses this feature but doesn't use *args in their callback. If accepted, we'll need to announce a backwards incompatible change.

mitar:
Hm, I see. Hmm.

mitar:
So what should we do here? Because this is really useful, because then you can use a function to have a dynamic/configurable value. Otherwise the function more or less cannot really decide much.

mitar:
Would it be OK to use [inspect](https://docs.python.org/2/library/inspect.html#inspect.getargspec) to inspect how many argument a function requires? To stay backwards compatible?

SeanHayes:
Hmm, I'd prefer not to use inspect.

Maybe instead we should add "for_list" to the Bundle class.

mitar:
> Maybe instead we should add "for_list" to the Bundle class.

Not sure what you mean here?

SeanHayes:
We could add an attribute to the Bundle class called "for_list". The bundle would get passed to your callback, and you'd be able to figure out if it's for a list or not.

mitar:
I see. Or we could just set `use_in` attribute, no?

PR #1271, opened 2015-02-09
More speed for dictionary builder

prosto-chewey:
Construction:
```python
{k: v for k, v in [('a', 1), ('b', 1), ('c', 1)]}
```
is faster than:
```python
dict((k, v) for k, v in [('a', 1), ('b', 1), ('c', 1)])
```

Simple timeit:
```python
In [1]: %timeit {k: v for k, v in [('a', 1), ('b', 1), ('c', 1)]}
1000000 loops, best of 3: 701 ns per loop

In [2]: %timeit dict((k, v) for k, v in [('a', 1), ('b', 1), ('c', 1)])
100000 loops, best of 3: 1.82
```

### DIFF
--- a/tastypie/serializers.py
+++ b/tastypie/serializers.py
@@ -237,9 +237,9 @@ class Serializer(object):
         if isinstance(data, (list, tuple)):
             return [self.to_simple(item, options) for item in data]
         if isinstance(data, dict):
-            return dict((key, self.to_simple(val, options)) for (key, val) in data.items())
+            return {key: self.to_simple(val, options) for (key, val) in data.items()}
         elif isinstance(data, Bundle):
-            return dict((key, self.to_simple(val, options)) for (key, val) in data.data.items())
+            return {key: self.to_simple(val, options)) for (key, val) in data.data.items()}
         elif hasattr(data, 'dehydrated_type'):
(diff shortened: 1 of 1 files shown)


georgedorn:
Travis reports multiple test failures in this branch.

tastypie/serializers.py
```diff
@@ -237,9 +237,9 @@ def to_simple(self, data, options):
         if isinstance(data, (list, tuple)):
             return [self.to_simple(item, options) for item in data]
         if isinstance(data, dict):
-            return dict((key, self.to_simple(val, options)) for (key, val) in data.items())
+            return {key: self.to_simple(val, options) for (key, val) in data.items()}
         elif isinstance(data, Bundle):
-            return dict((key, self.to_simple(val, options)) for (key, val) in data.data.items())
+            return {key: self.to_simple(val, options)) for (key, val) in data.data.items()}
```
SeanHayes: There's an extra parenthesis here.

SeanHayes:
Unfortunately, even once the extra parenthesis is removed, this won't work in Python 2.6. If support for that version is ever dropped this can be fixed and merged in.

SeanHayes:
Superseded by #1349

PR #1275, opened 2015-02-19
PUT response code correction

snopoke:
See commit for details: https://github.com/django-tastypie/django-tastypie/commit/abc0bef


georgedorn:
Good catch.  Thanks.

PR #1289, opened 2015-04-11
Fix/filter value to python

georgedorn:
Updating #785 to work with Django 1.6+.


tests/basic/tests/http.py
```diff
@@ -149,3 +149,10 @@ def test_private_cache_control(self):
 
         self.assertEqual(cache_control, set(["s-maxage=3600", "max-age=3600", "private"]))
         self.assertTrue('"johndoe"' in response.read().decode('utf-8'))
+
+    def test_filter_in(self):
+        connection = self.get_connection()
+        connection.request('GET', '/api/v1/notes/1/filterin/', headers={'Accept': 'application/json'})
```
georgedorn: I'd like to see this actually pass the filter in as part of the query, to show it splitting the csv list of IDs as well.

tests/basic/tests/views.py
```diff
@@ -258,3 +258,8 @@ def test_private_cache_control(self):
 
         self.assertEqual(cache_control, set(["s-maxage=3600", "max-age=3600", "private"]))
         self.assertTrue('"johndoe"' in response.content.decode('utf-8'))
+
+    def test_filter_in(self):
+        response = self.client.get('/api/v1/notes/1/filterin/', HTTP_ACCEPT='application/json')
```
georgedorn: In retrospect, I'm unhappy with this test.  It only shows that a view can call get_list with id__in a list, it doesn't actually show using the filter from the query parameters.

PR #1268, opened 2015-01-29
Add test for issue #1267, DateField is formatted as DateTimeField when using always_return_data = True

lsapan:
I opened issue #1267 because DateFields are formatted as DateTimeFields in the response when `always_return_data` is enabled on a resource. At first I thought it was only for PUT requests, but as the test shows, it actually is for POST requests as well. The test fails the assertion in the middle as well as the one at the end.

I added the test to the basic set in http.py. I apologize if this was not the appropriate location, the docs weren't explicit with where to put new tests.

PR #1302, opened 2015-05-28
fix incorrect error message when datefield

sgmart:
This fixes https://github.com/django-tastypie/django-tastypie/pull/1268

The issue was in https://github.com/django-tastypie/django-tastypie/blob/master/tastypie/resources.py#L1775, where the if statement consider both (`DateField` and `DateTimeField`), but because of tastypie also include a `DateField` inside `tastypie/fields.py`, is just enough to append an additional if to check for it and change the result.

### DIFF
--- a/tastypie/resources.py
+++ b/tastypie/resources.py
@@ -1793,8 +1793,10 @@ class BaseModelResource(Resource):
         result = default
         internal_type = f.get_internal_type()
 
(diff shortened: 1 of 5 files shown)


SeanHayes:
Duplicate of #1302

sgmart:
@SeanHayes, I've added a test for this issue.

SeanHayes:
Fixed by #1365

Issue #782, opened 2013-01-25
Provide URL of related resource in schema when type="related"


kennknowles:
Hello,

Please correct me if I have overlooked some options, but I believe that currently one cannot automatically build e.g. a Backbone-Relational client for a Tastypie API because related resources are not fully specified in the schema.

To be concrete, if I write code like this in the `my_api.resources` module.

```python
class FooResource(Resource):
    bar = ToOneField('my_api.resources.BarResource', attribute='bar')
    ...

class BarResource(Resource):
    ...
```

Then the resulting schema for `Foo` will include the following:

```
bar: {
   blank: false,
   default: "No default provided.",
   help_text: "A single related resource. Can be either a URI or set of nested resource data.",
   nullable: false,
   readonly: false,
   related_type: "to_one",
   type: "related",
   unique: false
}
```

It would be quite useful to also include a field like `resource_uri: /api/bar` (and also `schema_uri: /api/bar/schema` would be helpfully explicit rather than depending on Tastypie's URI structure). From browsing the code I believe this is as simple as showing something like `RelatedField.to_class().get_resource_uri()` but I would like to hear thoughts from core devs.

Thanks!

- Kenn

mnowotka:
What if related resource is not registered in the API? For example I only want to present B resource nested inside A resource so I create related field to B in A but not register B. This works fine but now there is no information about B fields/filtering/ordering in A schema. There should be an option to make a nested schema including all this information as well.

SeanHayes:
Reopening, mnowotka makes a good point about possibly needing embedded schemas for non-published related resources.

SeanHayes:
Fixed by #1309

SeanHayes:
Reopening because we need nested related schemas when related resources aren't URL accessible.

PR #1309, opened 2015-06-29
Provide URL of related resource in schema when type="related" (#782)

juditnovak:
…
Fix for issue #782 (https://github.com/django-tastypie/django-tastypie/issues/782)

Schema information only gives information about whether a related field is
'to_one' or 'to_many', but not about the actual resource type of the field. This
fix adds a new field, called 'related_schema' to the fields schema entry, which
defines the exact type of the related field:

    'user': {
        [..]
        'type': 'related'
        'related_type': 'to_one'
        'related_schema': '/api/v1/user/schema/'
        }

Documentation (schema example on the 'Interacting With The API' page) is updated accordingly.

### DIFF
--- a/tests/core/tests/resources.py
+++ b/tests/core/tests/resources.py
@@ -1349,7 +1349,7 @@ class CounterUpdateDetailResource(ModelResource):
 
 class ModelResourceTestCase(TestCase):
     fixtures = ['note_testdata.json']
-    urls = 'core.tests.field_urls'
+    urls = 'core.tests.resource_urls'
 
     def setUp(self):
         super(ModelResourceTestCase, self).setUp()
@@ -1696,7 +1696,7 @@ class ModelResourceTestCase(TestCase):
         self.assertEqual(resource.determine_format(request), 'application/xml')
 
     def test_build_schema(self):
-        related = RelatedNoteResource()
+        related = RelatedNoteResource(api_name='v1')
         schema = adjust_schema(related.build_schema())
         expected_schema = {
             'filtering': {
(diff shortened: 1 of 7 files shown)


georgedorn:
The unit tests for schema generation will need to be updated before this can be merged.  See the travis build for more details.

mikebryant:
@georgedorn Unit tests updated :)

mikebryant:
@SeanHayes Rebased.
Not sure why the coverage is dropping - tests/core/tests/field_urls.py is still used in tests/core/tests/fields.py

mikebryant:
@SeanHayes What needs to be done to progress this?

tests/core/tests/resources.py
```diff
@@ -1349,7 +1349,7 @@ class Meta:
 
 class ModelResourceTestCase(TestCase):
     fixtures = ['note_testdata.json']
-    urls = 'core.tests.field_urls'
+    urls = 'core.tests.resource_urls'
```
SeanHayes: I think that because of this line, core.tests.field_urls now has no coverage, which is why the percentage dropped.

core.tests.field_urls is referenced here too: https://coveralls.io/files/1857577512#L955
But it doesn't look like it uses the test client at all.

Since it appears field_urls.py isn't actually needed, I would delete it, and then delete this line:
https://coveralls.io/files/1857577512#L957

Don't forget to add this PR to the list in docs/release_notes/dev/rst. Then squash your commits.

mikebryant:
@SeanHayes field_urls removed, reference removed.
Added release notes.
Squashed the commits.

Coverage is now only affected by the removal of the line in `tests/core/tests/fields.py`, dropping the percentage. (Number of missed lines has not increased)

SeanHayes:
Thanks!

PR #1318, opened 2015-07-16
fixes Django 1.8 regression for saving related fields

w1zeman1p:
Before the django 1.8 upgrade there was a check in resources.py for the
existence of the related obj when saving a bundle. This check was
removed but we still had a few cases where that related obj might be
`None`. This re-adds the condition, and is forwards compatible with
django 1.8.


georgedorn:
As travis indicates, this breaks the build.  We also need a test for the case you are fixing.

PR #1320, opened 2015-08-02
Add some debug messages to certain cases.

fluffels:
Previously, only trace backs were provided in the case that:
- a `resource_uri` could not be inferred from a `Model` instance, or
- an operation was rejected due to lack of authorization.

This PR adds debug messages for these two cases, which would hopefully help during development.


SeanHayes:
The NoReverseMatch logging seems helpful, I'm just not sure what the unauthorized logging is needed for since a 4xx response will get returned. Could you elaborate?

fluffels:
If I recall correctly, I wanted to know which Authorization class was responsible for rejecting the request.

SeanHayes:
Gotcha. Yeah, the Unauthorized exception doesn't carry a lot of info, we should probably fix that by setting some attributes when it's raised.

fluffels:
That would be nice. The reason I wanted logging though was because sometimes it is nice to have server-side records of issues in cases where a third party is consuming the API and they run into an issue, or someone experiences a problem with your website and they don't write down the error message.

fluffels:
GitHub doing GitHub things...

georgedorn:
This seems to have no content anymore.  I'm happy to include this if it wasn't included elsewhere; reopen or open a new PR if you are still interested in implementing this.

PR #1321, opened 2015-08-07
Add custom_api decorator which helps to implement non-CRUD end point

yfli:


SeanHayes:
I think I prefer the approach this PR takes, although that needs some cleanup too: #724

yfli:
@SeanHayes I was busy and thinking to add the missed unit test recently. Could you specify what the cleanups? I could add them together.

SeanHayes:
Sorry, I was referring to the other PR. I'm not sure what we'll go with, I think a design decision is needed.

PR #1322, opened 2015-08-12
[doc] add missing imports in non-orm example

nikolas:
The example is still missing some but it's nice to show where `Resource` can be imported from.


georgedorn:
Yeah, this makes things clearer  I approve.

PR #1329, opened 2015-09-05
GIS tests with spatialite

SeanHayes:


SeanHayes:
Not sure why, but Spatialite seems to be working in Python 3 out of the box, without installing pyspatialite.

All but 2 tests work in Spatialite, which are skipped using a decorator called skipIfSpatialite. Still need to configure PostGIS testing.

PR #1330, opened 2015-09-05
Reduces a query when authing users via api keys.

SeanHayes:


georgedorn:
We appear to have a nondeterministic test, probably pertaining to the throttling code relying on timing or something similar.

SeanHayes:
@georgedorn Should be fixed now.

PR #1332, opened 2015-09-07
RemovedInDjango19Warning: django.utils.importlib will be removed in Django 1.9.

jelenak:
Let's do not show this notification:

```
/env/src/django-tastypie/tastypie/fields.py:7: RemovedInDjango19Warning: django.utils.importlib will be removed in Django 1.9.
  from django.utils import datetime_safe, importlib
```


SeanHayes:
Thanks!

PR #1338, opened 2015-09-09
Unquote url-s containing escaped characters

rkojedzinszky:
This would allow processing resource_uri's like /~krichy/api/...


SeanHayes:
I tried adding more tests for get_via_uri() and it seems to resolve URIs with tildes fine.

Issue #629, opened 2012-08-17
Make the format not implemented error more descriptive


jleclanche:
Hi, I recently hit the "format not implemented" error as I'm sure many people have. Given that the [explanation](http://django-tastypie.readthedocs.org/en/v0.9.11/debugging.html#i-m-getting-xml-output-in-my-browser-but-i-want-json-output) for it is non-obvious, I suggested on IRC making the error more descriptive, the likes of:

```"Sorry, support for <dump accept header here> is not implemented yet. Please append "?format=json" to your URL."```

Just a thought.

**Comment by SiggyF** [comment]
I also get this error message if I use curl and add text/html as an accepted mime type. 
I would suggest to remove the [http](https://github.com/django-tastypie/django-tastypie/blob/master/tastypie/serializers.py#L86) from the list of supported formats. The list of supported formats is passed to the [best_match](https://github.com/dbtsai/python-mimeparse/blob/master/mimeparse.py#L137) function. Because tastypie passes `text/html` to the list of supported formats it will return `text/html` as the preferred format (because it has the default [q=1](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html)), but it is in fact not supported (yet). 

This is an example of a curl request that fails, even though it supports json:
```
curl -H "Accept: application/json;q=0.9,text/html" http://localhost:8000/api/v1/
```

I'd be happy to put this in a pull request if that would work for you?

SeanHayes:
@SiggyF That would be great! Make sure to add tests.

SeanHayes:
The related PR has been merged. Another PR including a more descriptive error message would still be welcome, if you're interested.

PR #1343, opened 2015-09-11
remove unimplemented to_html/from_html

SiggyF:
This pull request is a followup on #629. 
I have removed the unimplemented `to_html` and `from_html`. 
This issue solves the problem that text/html is returned where a application/xml or application/json is requested. See details in #629.

I updated tests and docs and added 2 new tests (with Accept header from Chrome/Firefox). The pull request was tested without errors with tox with environments py27, 1.7 and 1.8, py3.4, 1.7 and 1.8.

Please consider incorporating this pull request. 

I also considered implementing the to_html/from_html methods. I could not find a good reference/example where html is used as a serialisation format. For the `to_html` I found the [`_repr_html_`](https://ipython.org/ipython-doc/dev/config/integrating.html#rich-display) in IPython which is appealing, but it's not a data format. Alternatives that one could consider are the [microdata](https://html.spec.whatwg.org/multipage/microdata.html) format (not clear how to_html would look like) and the [data-](http://www.w3.org/TR/html5/dom.html#embedding-custom-non-visible-data-with-the-data-*-attributes) attributes. Comments in the code suggested that perhaps the intention was to implement something like a `from_request` method, where a form post request could be used as input.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -88,6 +88,7 @@ Contributors:
 * Renjith Thankachan for a patch that fixes `django.utils.importlib` error in Django 1.9+ release.
 * Sean Hayes (SeanHayes) for test suite and CI enhancements, various patches
 * Michael Thornhill (mthornhill) for strict dict checking bugfix
+* Fedor Baart (siggyf) for an update on the serializer
 
 Thanks to Tav for providing validate_jsonp.py, placed in public domain.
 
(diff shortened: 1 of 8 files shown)


SeanHayes:
Thanks!

IMO supporting HTML doesn't really make sense. Eventually I'd like to have a JS CRUD frontend for Tastypie though using something like Backbone or Angular.

PR #1344, opened 2015-09-14
Update fields.rst to include a warning about saving related objects.

strets123:


SeanHayes:
Could you show us the code that's giving you this problem? Or better yet, a test case? I would think that if the ID and resource_uri are included on the data you POST back, Tastypie ought to look up the existing model and merge the data.

strets123:
I will try to do a specific test case this week. I am not disputing that Tastypie looks up the existing model, it does. The issue is that if a field is in Meta.excluded then it is assumed to be the default value. Therefore password gets set to an empty string.

SeanHayes:
Yeah, that's something we'll want to get fixed. Thanks for the test case!

strets123:
See pull request for a failing test case which demonstrates when this bug happens. Also added a fix for the socket exceptions.

SeanHayes:
Thanks! I'll pull these changes into another PR with a fix for the test failure.

strets123:
Thanks Sean, glad to help out.

SeanHayes:
I found the problem.

fields.py:612
if unique_keys and fk_resource.can_update():

can_update() returns False if PUT isn't an allowed method.

The real bug is that the related resource got saved when updating isn't allowed. I'm working on a fix.

PR #1346, opened 2015-09-16
Removed obsolete Django 1.3/1.4 compatibility code (current minimum …

SeanHayes:
…supported version is 1.5).


georgedorn:
We technically could stop running tests against 1.6 as well - that version
is essentially EOL.

SeanHayes:
I'd be in favor of that, though that would involve even more changes, such as removing South migrations.

SeanHayes:
@georgedorn If you're OK with this PR, we can merge it in, then I'll get started on a new PR removing Django 1.5 and 1.6 support.

georgedorn:
Oh, yeah, :+1:

PR #1349, opened 2015-09-16
WIP: Removed Django 1.5-1.6 support.

SeanHayes:
Still need to update the docs, remove references to South, etc. Some of the compat.py and timezone stuff can probably be cleaned up.


SeanHayes:
I think before this gets merged we should make another pip release so recent bug fixes are available for older Django versions, then we should bump the Tastypie version number when this is merged.

SeanHayes:
Going to open a new PR into a new release branch.

PR #1357, opened 2015-10-03
1353 Debugging clarity when passing not Resource instance

annacorobco:


SeanHayes:
Squashed commits into #1361.

Thanks!

PR #1363, opened 2015-10-09
Fix hydrating/saving of related resources.

SeanHayes:
Related to #1344 

Some test cases have been split up into separate test case to make debugging easier, and some new tests and assertions were added, but otherwise the same tests are being executed.

Notable changes:
1. We skip saving resources loaded via a URI (RelatedField().resource_from_uri()) since we know there's no changes from the client.
2. In RelatedField().resource_from_data(), we no longer check can_update(). We don't use the can_* methods anywhere else, and those methods don't check permissions, just allowed HTTP methods. Permissions checking still happens in Resource().save().
3. In RelatedField().resource_from_data(), we no longer call Resource().obj_update(). It seemed really inappropriate to be saving anything here (during the hydration step), my guess is this code got checked in without anyone taking too close a look; no permissions checks were being done (!!!), just an HTTP method check (can_update) on the related resource.

### DIFF
--- a/tastypie/bundle.py
+++ b/tastypie/bundle.py
@@ -19,6 +19,7 @@ class Bundle(object):
                  related_name=None,
                  objects_saved=None,
                  related_objects_to_save=None,
+                 via_uri=False,
                  ):
         self.obj = obj
         self.data = data or {}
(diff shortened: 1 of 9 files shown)


SeanHayes:
@georgedorn You might want to take a look at this. I'll go ahead and merge in a few days if you're not able to get around to it.

PR #1373, opened 2015-10-11
Return 400 error when request body has malformed syntax. (based on PR #649 by piotrbulinski)

SeanHayes:


SeanHayes:
Merging to a different dranch in #1375

PR #1376, opened 2015-10-12
When Tastypie handles a 500 error, log the exception and send the signal…

SeanHayes:
… before before returning a debug response.


SeanHayes:
I think coverage went down because _handle_500 has less code in it.

tastypie/resources.py
```diff
@@ -264,36 +264,28 @@ def _handle_500(self, request, exception):
...
-        NOT_FOUND_EXCEPTIONS = (NotFound, ObjectDoesNotExist, Http404)
-
-        if isinstance(exception, NOT_FOUND_EXCEPTIONS):
+        if isinstance(exception, (NotFound, ObjectDoesNotExist, Http404)):
             response_class = HttpResponseNotFound
-            response_code = 404
+
+        log = logging.getLogger('django.request.tastypie')
+        log.error('Internal Server Error: %s' % request.path, exc_info=True,
+                  extra={'status_code': response_class.status_code, 'request': request})
```
clayg: THIS WOULD BE GREAT!

PR #1378, opened 2015-10-12
Fixes Reverse One to One Relationships (Replaces #568)

SeanHayes:


tastypie/fields.py
```diff
@@ -763,6 +764,15 @@ def hydrate(self, bundle):
...
             return value
 
+        if bundle.obj and isinstance(getattr(bundle.obj.__class__, self.attribute), SingleRelatedObjectDescriptor):
+            # This is the case when we are writing to a reverse one to one field.
+            # Enable related name to make this work fantastically.
+            # see https://code.djangoproject.com/ticket/18638 (bug; closed; worksforme)
+            # and https://github.com/toastdriven/django-tastypie/issues/566
+
+            # this gets the related_name of the one to one field of our model
+            self.related_name = getattr(bundle.obj.__class__, self.attribute).related.field.name
```
dotsbb: @SeanHayes You are modifying field instance based on bundle's data. Take into account that some related field could have `null=True`

SeanHayes: Even if null=True I think the field instance should still be there. I should change it to not modify the field instance based on bundle data though, related_name should be set during __init__.

Opened #1415

tastypie/fields.py
```diff
@@ -763,6 +764,15 @@ def hydrate(self, bundle):
         if value is None:
             return value
 
+        if bundle.obj and isinstance(getattr(bundle.obj.__class__, self.attribute), SingleRelatedObjectDescriptor):
```
prokaktus: If `self.attribute` is a composite attribute (`country__president`), then we get exception here, because our class doesn't have the attribute `country__president`, but have the attribute `country`, which have the attribute `president`). And it's taking into account in `self._attrs`:
```
self._attrs = attribute.split('__') if attribute is not None and isinstance(attribute, six.string_types) else []
```
Maybe I should open issue + test on this case?

SeanHayes: Yeah, that'd be very helpful.

SeanHayes: See #1446

prokaktus: @SeanHayes oh, thanks! I'll take a look at new pypi version as soon as I can! Great work and many thanks!

Issue #1323, opened 2015-08-16
Put / patch requests on non PK data


nicerworld:
I am using detail_uri_field as follows to get rid of IDs out of data, I need business growth to be a secret. This works great on GET requests...

class Meta:
	detail_uri_name='uuid'
	def prepend_urls(self):
		return [
			url(r'^(?P<resource_name>%s)/(?P<uuid>[\w\.-]+)/$' % self._meta.resource_name, self.wrap_view('dispatch_detail'), name='api_dispatch_detail'),
		]

But on PATCH/ PUT I get a duplicate error, i.e, it's doing a create rather than update  It looks to me as though the uuid doesn't get used to get the object that needs updating, is this true? I am in the source but hoping someone can help me solve this the right way!
Thanks
Rich

nicerworld:
In addition, when I do a get request on  /api/v1/user/fb2fe716ca03413496a58cc8868c0fd5/?format=json

I get back data with no resource URI. I tried adding one matching above, I even tried adding an ID field to the JSON posted with the actual ID of the object to see if it would give me a clue, but it's still trying to create rather than update.

SeanHayes:
I think specifying Meta.detail_uri_name should be all you need to do. The regex in your prepend_urls doesn't look quite right.

Could you post your models and resources? If you're still having trouble I can put together a test case and try to fix the issue.

SeanHayes:
Note to self/devs: need more tests where detail_uri_name gets customized.

burnhambp:
Hi Sean,

I moved onto a different area of project but I am back to it now and could really use a fix! 

I can confirm this is definitely an issue. Doing a put request to the resource_uri returned on a resource is creating a new record. 

I think someone else had this issue as well (http://newscentral.exsees.com/item/47beeac8c38dde2a7bd948472a007f39-a9afeea5ae05fd8dbf1551104be330b4)

Here is my resource:

class BusinessAreaResource(ModelResource):
	class Meta:
		detail_uri_name='uuid'
		queryset = BusinessArea.objects.all()
		resource_name = 'business_area'
		always_return_data = True
		authentication = Authentication() 
		authorization = Authorization()
		excludes=['id', 'created']

	def override_urls(self):
		return [
			url(r'^(?P<resource_name>%s)/(?P<uuid>[\w\d_.-]+)/$' % self._meta.resource_name, self.wrap_view('dispatch_detail'), name='api_dispatch_detail'),
		]

Here is the model:

[...]

PR #1387, opened 2015-10-20
Fixes bug which occurs when detail_uri_name field has a default value (Issue #1323)

SeanHayes:


tastypie/resources.py
```diff
@@ -705,7 +709,7 @@ def build_bundle(self, obj=None, data=None, request=None, objects_saved=None, vi
 
         If no object is provided, an empty object from
         ``Resource._meta.object_class`` is created so that attempts to access
-        ``bundle.obj`` do not fail.
+        ``bundle.obj`` do not fail. You can prevent this by passing in `obj=False`.
```
SeanHayes: This feels sort of wrong, need to come up with better solution.

PR #1401, opened 2015-12-07
Fixed Authorization

marcelometal:
Error: 'NoneType' object has no attribute 'has_perm'.

### DIFF
--- a/tastypie/authorization.py
+++ b/tastypie/authorization.py
@@ -179,7 +179,7 @@ class DjangoAuthorization(Authorization):
             get_module_name(klass._meta)
         )
 
-        if not bundle.request.user.has_perm(permission):
+        if not bundle.request.user or not bundle.request.user.has_perm(permission):
             return []
 
         return object_list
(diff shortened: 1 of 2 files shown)


SeanHayes:
Just wondering, normally request.user will either be a User or AnonymousUser. What situations is it set to None?

marcelometal:
I'm using django-tastypie-oauth and django-oauth-toolkit. It happens when a 'client_credentials' is used to POST a resource URL.

Please, take a look at: [django-oauth-toolkit](https://github.com/evonove/django-oauth-toolkit/blob/772dbed1ccc6a828fc85466a1bf9043f78b18385/oauth2_provider/oauth2_validators.py#L307).

SeanHayes:
Thanks for getting back to me.

According to Django it's always going to be User or AnonymousUser: https://docs.djangoproject.com/en/1.9/ref/request-response/#django.http.HttpRequest.user

I'd rather not complicate our code or add any workaround for bugs in other packages if I can avoid it. I'd recommend opening a pull request for those other projects to correctly set an instance of AnonymousUser. You can also write your own Authorization class that does the same thing as this PR.

PR #1402, opened 2015-12-08
stop using django.conf.urls.patterns

brad:
Stop using `django.conf.url.patterns`, which will be removed in Django 1.10.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -71,17 +71,17 @@ A basic example looks like:
 
     # urls.py
     # =======
-    from django.conf.urls.defaults import *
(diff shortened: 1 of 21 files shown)


SeanHayes:
Awesome, thanks!

This looks just about ready to merge. The only issue is this is a backwards incompatible change for anyone who uses Tastypie and has overridden URLs, so this will need to be in the next minor (not micro) release, and we'll have to add a note to the changelogs for that release.

SeanHayes:
Actually, nevermind, patterns() just returns a regular list, so everything should be fine.

Issue #1407, opened 2016-01-06
Need a much safer default: `DjangoAuthorization` should not allow `read` access to all model object


thomasyip:
Looking at `tastypie/authorization.py` (around like 133 -- master branch on 1/4/2016), the default, both `read_list` and `read_details` bypass `user.has_perm()` check, which is quite unsafe and is a bad default.

Django's Admin default is unconventional. So, I could see how it was misinterpreted.

https://docs.djangoproject.com/es/1.9/topics/auth/default/#permissions-and-authorization

    The Django admin site uses permissions as follows:
    *   ... 
    * Access to view the change list, view the “change” form and change an object is limited to users with the “change” permission for that type of object.

Essentially, "change_xyz" is the permission code for both "read" and "update". I think the better default would be following Django's Admin:

```python
diff --git a/tastypie/authorization.py b/tastypie/authorization.py
index 1d6f5aa..44b2d56 100644
--- a/tastypie/authorization.py
+++ b/tastypie/authorization.py
@@ -151,22 +151,14 @@ class DjangoAuthorization(Authorization):
         return model_klass
 
     def read_list(self, object_list, bundle):
-        klass = self.base_checks(bundle.request, object_list.model)
-
-        if klass is False:
-            return []
+        # By default, follows `ModelAdmin` "convention" to use `app.change_model`
+        # `django.contrib.auth.models.Permission` for both viewing and updating.
+        # https://docs.djangoproject.com/es/1.9/topics/auth/default/#permissions-and-authorization
 
-        # GET-style methods are always allowed.
-        return object_list
+        return self.update_list(object_list, bundle)
 
     def read_detail(self, object_list, bundle):
-        klass = self.base_checks(bundle.request, bundle.obj.__class__)
-
-        if klass is False:
-            raise Unauthorized("You are not allowed to access that resource.")
-
-        # GET-style methods are always allowed.
-        return True
+        return self.update_detail(object_list, bundle)
 
```

miraculixx:
I appreciate this is merged already but I still take issue with it. Before opening a new request, here's my concern:

This change unfortunately breaks existing code and thus is backwards incompatible (previously all GET requests would pass DjangoAuthorization). To fix one's code, the only two options are to either give all users 'change' permission (bad), or to subclass DjangoAuthorization for a custom implementation of the read actions (potentially lots of work). Was this intended?

[...]

Issue #1408, opened 2016-01-06
Add `DjangoObjectAuthorization`


thomasyip:
Since, at least Django 1.5 (maybe earlier), its auth package allow permission check for specific object.

https://docs.djangoproject.com/en/1.5/ref/contrib/auth/#django.contrib.auth.models.User.has_perm
https://docs.djangoproject.com/en/1.9/ref/contrib/auth/#django.contrib.auth.models.User.has_perm

The ability can be provided by `AUTHENTICATION_BACKENDS`, a few implementations are readily available. Such as `django-trusts` (shameless plug) and `django-guardian`.

I think Tastypie is the right place to provide `DjangoObjectAuthorization` "out of the box".
# 

Adding the TODO items from #1409 that is related to DjangoObjectAuthorization here.
1. Add tests for DjangoObjectAuthorization.
2. Add docs for DjangoObjectAuthorization.
3. Mention this PR in docs/release_notes.dev.rst

Thanks, @SeanHayes

thomasyip:
Once PR #1409 is merged, one can add DjangoObjectAuthorization with a single method override, in case anyone need it soon:


```python
class DjangoObjectAuthorization(DjangoAuthorization):
    '''
    Uses permission checking from ``django.contrib.auth`` to map
    ``POST / PUT / DELETE / PATCH`` to their equivalent Django auth
    permissions.
    Both the list & detail simply check the object they're based on.
    Object level authorization api is added since django 1.5. However,
    it will default to no-access unless an AUTHENTICATION_BACKENDS is
    setup to handle those checks.
    '''

    def check_user_perm(self, user, permission, obj_or_list):
        return user.has_perm(permission, obj_or_list)
```

thomasyip:
Closing 4 years old issue, because it was merged. See PR #1409.

PR #1409, opened 2016-01-06
Issue #1407: better default for DjangoAuthorization.

thomasyip:
Fix for Issue #1407, better default for DjangoAuthorization.

Looking at `tastypie/authorization.py` (around like 133 -- master branch on 1/4/2016), the default, both `read_list` and `read_details` bypass `user.has_perm()` check, which is quite unsafe and is a bad default.

Django's Admin default is unconventional. So, I could see how it was misinterpreted.

https://docs.djangoproject.com/es/1.9/topics/auth/default/#permissions-and-authorization

    The Django admin site uses permissions as follows:
    *   ... 
    * Access to view the change list, view the “change” form and change an object is limited to users with the “change” permission for that type of object.

Essentially, "change_xyz" is the permission code for both "read" and "update". I think the better default would be following Django's Admin:

### DIFF
--- a/tastypie/authorization.py
+++ b/tastypie/authorization.py
@@ -1,7 +1,5 @@
 from __future__ import unicode_literals
 
-import settings
-
 from tastypie.exceptions import TastypieError, Unauthorized
 from tastypie.compat import get_module_name
 
--- a/tastypie/authorization.py
+++ b/tastypie/authorization.py
@@ -141,140 +141,11 @@ class DjangoAuthorization(Authorization):
     Both the list & detail variants simply check the model they're based
     on, as that's all the more granular Django's permission setup gets.
     """
-    def base_checks(self, request, model_klass):
-        # If it doesn't look like a model, we can't check permissions.
-        if not model_klass or not getattr(model_klass, '_meta', None):
-            return False
 
-        # User must be logged in to check permissions.
-        if not hasattr(request, 'user'):
-            return False
-
-        return model_klass
-
-    def read_list(self, object_list, bundle):
-        # By default, follows `ModelAdmin` "convention" to use `app.change_model`
-        # `django.contrib.auth.models.Permission` for both viewing and updating.
-        # https://docs.djangoproject.com/es/1.9/topics/auth/default/#permissions-and-authorization
-
-        return self.update_list(object_list, bundle)
-
-    def read_detail(self, object_list, bundle):
-        return self.update_detail(object_list, bundle)
-
-    def create_list(self, object_list, bundle):
-        klass = self.base_checks(bundle.request, object_list.model)
-
-        if klass is False:
-            return []
-
-        permission = '%s.add_%s' % (
-            klass._meta.app_label,
-            get_module_name(klass._meta)
-        )
-
-        if not bundle.request.user.has_perm(permission):
-            return []
-
-        return object_list
-
-    def create_detail(self, object_list, bundle):
-        klass = self.base_checks(bundle.request, bundle.obj.__class__)
-
-        if klass is False:
-            raise Unauthorized("You are not allowed to access that resource.")
-
-        permission = '%s.add_%s' % (
-            klass._meta.app_label,
-            get_module_name(klass._meta)
-        )
-
-        if not bundle.request.user.has_perm(permission):
-            raise Unauthorized("You are not allowed to access that resource.")
-
-        return True
-
-    def update_list(self, object_list, bundle):
-        klass = self.base_checks(bundle.request, object_list.model)
-
-        if klass is False:
-            return []
-
-        permission = '%s.change_%s' % (
-            klass._meta.app_label,
-            get_module_name(klass._meta)
-        )
-
-        if not bundle.request.user.has_perm(permission):
-            return []
-
-        return object_list
-
-    def update_detail(self, object_list, bundle):
-        klass = self.base_checks(bundle.request, bundle.obj.__class__)
-
-        if klass is False:
-            raise Unauthorized("You are not allowed to access that resource.")
-
-        permission = '%s.change_%s' % (
-            klass._meta.app_label,
-            get_module_name(klass._meta)
-        )
-
-        if not bundle.request.user.has_perm(permission):
-            raise Unauthorized("You are not allowed to access that resource.")
-
-        return True
-
-    def delete_list(self, object_list, bundle):
-        klass = self.base_checks(bundle.request, object_list.model)
-
-        if klass is False:
-            return []
-
-        permission = '%s.delete_%s' % (
-            klass._meta.app_label,
-            get_module_name(klass._meta)
-        )
-
-        if not bundle.request.user.has_perm(permission):
-            return []
-
-        return object_list
-
-    def delete_detail(self, object_list, bundle):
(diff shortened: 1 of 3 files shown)


thomasyip:
The patch failed those test that assumes everyone has read-access. I could patch the tests, but I think tastypie team need to decide whether it is the directly they want to go first.

tastypie/authorization.py
```diff
@@ -1,5 +1,7 @@
 from __future__ import unicode_literals
 
+import settings
```
SeanHayes: Should be django.conf.settings

tastypie/authorization.py
```diff
@@ -263,3 +257,88 @@ def delete_detail(self, object_list, bundle):
...
+    Both the list & detail simply check the object they're based on.
+    Object level authorization api is added since django 1.5. However,
+    it will default to no-access unless an AUTHENTICATION_BACKENDS is
+    setup to handle those checks.
+    '''
+
+    # By default, follows `ModelAdmin` "convention" to use `app.change_model`
+    # `django.contrib.auth.models.Permission` for both viewing and updating.
+    # https://docs.djangoproject.com/es/1.9/topics/auth/default/#permissions-and-authorization
+    READ_PERM_CODE = getattr(settings, 'TASTYPIE_READ_PERM_CODE', 'change')
```
SeanHayes: Could you just hard code this to 'change' here, instead of adding a new setting? If people really need to change it they can subclass and override READ_PERM_CODE.

thomasyip: What you proposed makes sense. I will add it a patch (and `settings` to django...)

If you think it makes sense to include `DjangoObjectAuthorization`, would you like me to collapse the two classes. The `DjangoObjectAuthorization` and `DjangoAuthorization` really only different in a single method call, namely `user.has_perm(permission, obj_or_list)` vs `user.has_perm(permission)`.

Something like the code below. 

(I am in the process of updating my own app to Django 1.8 (yeah, a bit late to the party) and in a bit of a git mess. Will update the pull request once I get back to the stage that I can run test. Sorry that I might not able to update this PR until sometime next week. Sorry about that)

```python
class DjangoAuthorization(Authorization):
    """
    Uses permission checking from ``django.contrib.auth`` to map
    ``POST / PUT / DELETE / PATCH`` to their equivalent Django auth
    permissions.

    Both the list & detail variants simply check the model they're based
    on, as that's all the more granular Django's permission setup gets.
    """

    # By default, following `ModelAdmin` "convention", `app.change_model` is used
    # `django.contrib.auth.models.Permission` as perm code for viewing and updating.
    # https://docs.djangoproject.com/es/1.9/topics/auth/default/#permissions-and-authorization
    READ_PERM_CODE = getattr(settings, 'TASTYPIE_READ_PERM_CODE', 'change')

    def base_checks(self, request, model_klass):
        # If it doesn't look like a model, we can't check permissions.
        if not model_klass or not getattr(model_klass, '_meta', None):
            return False

        # User must be logged in to check permissions.
        if not hasattr(request, 'user'):
            return False

        return model_klass

    def check_user_perm(self, user, permission, obj_or_list):
        return user.has_perm(permission)

    def perm_list_checks(self, request, code, obj_list):
        klass = self.base_checks(request, obj_list.model)
        if klass is False:
            return []

        permission = '%s.%s_%s' % (
            klass._meta.app_label,
            code,
            get_module_name(klass._meta)
        )

        if self.check_user_perm(request.user, permission, obj_list):
            return obj_list

        return obj_list.none()

    def perm_obj_checks(self, request, code, obj):
        klass = self.base_checks(request, obj.__class__)
        if klass is False:
            raise Unauthorized("You are not allowed to access that resource.")

        permission = '%s.%s_%s' % (
            klass._meta.app_label,
            code,
            get_module_name(klass._meta)
        )

        if request.user.has_perm(permission, obj):
            return True

        return False

    def read_list(self, object_list, bundle):
        return self.perm_list_checks(bundle.request, self.READ_PERM_CODE, object_list)

    def read_detail(self, object_list, bundle):
        return self.perm_obj_checks(bundle.request, self.READ_PERM_CODE, bundle.obj)

    def create_list(self, object_list, bundle):
        return self.perm_list_checks(bundle.request, 'add', object_list)

    def create_detail(self, object_list, bundle):
        return self.perm_obj_checks(bundle.request, 'add', bundle.obj)

    def update_list(self, object_list, bundle):
        return self.perm_list_checks(bundle.request, 'change', object_list)

    def update_detail(self, object_list, bundle):
        return self.perm_obj_checks(bundle.request, 'change', bundle.obj)

    def delete_list(self, object_list, bundle):
        return self.perm_list_checks(bundle.request, 'delete', object_list)

    def delete_detail(self, object_list, bundle):
        return self.perm_obj_checks(bundle.request, 'delete', bundle.obj)

class DjangoObjectAuthorization(DjangoAuthorization):
    '''
    Uses permission checking from ``django.contrib.auth`` to map
    ``POST / PUT / DELETE / PATCH`` to their equivalent Django auth
    permissions.

    Both the list & detail simply check the object they're based on.
    Object level authorization api is added since django 1.5. However,
    it will default to no-access unless an AUTHENTICATION_BACKENDS is
    setup to handle those checks.
    '''

    def check_user_perm(self, user, permission, obj_or_list):
        return user.has_perm(permission, obj_or_list)
```

SeanHayes: Yeah, that'd be great, more code reuse is better. Just make sure to:
1. Change READ_PERM_CODE = getattr(settings, 'TASTYPIE_READ_PERM_CODE', 'change') to READ_PERM_CODE = 'change'
2. Add tests for DjangoObjectAuthorization.
3. Fix tests for DjangoAuthorization (verify that reading without change permissions fails, verify that reading with change permissions succeeds)
4. I think the current docs warn that DjangoAuthorization always allows read permissions, that should be updated.
5. Add docs for DjangoObjectAuthorization.
6. Mention this PR in docs/release_notes.dev.rst

thomasyip: How about adding a line to BACKWARDS-INCOMPATIBLE.txt?

SeanHayes: Ahh, right, do that too.

SeanHayes: Actually, that file hasn't been updated in over 3 years, I wouldn't worry about it. Since then we've just been adding warnings to the release notes: e.g.  http://django-tastypie.readthedocs.org/en/latest/release_notes/v0.12.0.html

thomasyip: 👌

thomasyip:
Hi @SeanHayes, I made the said change related to issue #1407. 

I looked at the docs, couldn't find reference about "read-only". I read `authorization.rst` a few times by eyes, and grep all occurrences of "read" and "DjangoAuthorization" under `docs/` and didn't see a mention. I think we are cleared.

I removed code related to #1408 (i.e., DjangoObjectAuthorization) altogether. It is a bit more involving, and I will make another PR when I got chance.

I feel that it should not delay #1407 (i.e., Safer default).

I feel that it PR is ready to go. Let me know if you see anything else.

Thank a lot.


(btw, thanks for the list of TODO item, after two weeks, it really help refresh the memory)

SeanHayes:
Merged into master.

thomasyip:
Thank you!

PR #1414, opened 2016-01-15
Added error message when JSON requests cannot be deserialized

jlucas91:
As it currently stands, if a user makes a POST/PUT request with invalidly formatted JSON they receive a response that looks something like this like this:

`{"error":""}`

which, of course, isn't very helpful! This change provides a human readable error message when invalidly formatted JSON is submitted. 

@SeanHayes Hope this helps - let me know if there's anything I can do to improve it.


SeanHayes:
Thanks!

Issue #1411, opened 2016-01-12
Fix assertHttpAccepted


SeanHayes:
Bug was introduced in #1054

SeanHayes:
This bug was introduced here: https://github.com/django-tastypie/django-tastypie/pull/1054/files#diff-52d6df9d1b52221accfe27f100877d67R331

The assertion was put in the wrong place, it should have been in one of the tests in tests/core/tests/, but instead they put it in the test case we provide. The correct fix is to remove that line and add an assertion that the header doesn't exist in our test suite.

PR #1416, opened 2016-01-15
Split assertHttpAccepted

mattbriancon:
Pull out the HTTP No Content assert into a separate method, `assertHttpNoContent`.

Fixes https://github.com/django-tastypie/django-tastypie/issues/1411


mattbriancon:
It's worth noting that this will break tests that are currently using `assertHttpAccepted` with responses that have status code 204. We could add back the check for both 202 and 204 and just leave `assertHttpNoContent` as an upgrade path.

SeanHayes:
I updated #1411 with a description of the correct fix.

mattbriancon:
@SeanHayes updated per your recommendation.

SeanHayes:
Thanks!

Issue #1421, opened 2016-01-19
__repr__ should return str in Py2


prokaktus:
Hi all! Thanks for maintaining project!

`__repr__` in bundle should return str function in Py2. It's standard python-behavior. And some programs may rely on that fact.
https://docs.python.org/2/reference/datamodel.html#object.__repr__

Also, I have problem in one Django-project with representation of bundles. My bundle.data can contain dictionary with models. Models have defined `__unicode__`. But Django under the hood convert `__unicode__`-result to ascii in `__repr__` and `__str__` functions. And calling `__repr__` inside dictionary while representing bundle caused `UnicodeDecodeError`.

Example:
```
bundle.data = [
    {
        'model': my_model  # repr(my_model) will return str, not unicode
    }
]

repr(bundle) # calling __repr__ for my_model, it's Python's behavior
             # UnicodeDecodeError
```

Thanks!

PR #1422, opened 2016-01-19
Changed repr value for Bundle to str in PY2

prokaktus:
Attach pull-request for issue #1421.

I'll be glad to listen any objections or suggestions! Thanks for your time!

### DIFF
--- a/tastypie/bundle.py
+++ b/tastypie/bundle.py
@@ -1,5 +1,6 @@
 from __future__ import unicode_literals
 from django.http import HttpRequest
+from django.utils import six
 
 
 # In a separate file to avoid circular imports...
@@ -32,4 +33,7 @@ class Bundle(object):
         self.via_uri = via_uri
 
     def __repr__(self):
-        return "<Bundle for obj: '%s' and with data: '%s'>" % (self.obj, self.data)
+        repr_string = "<Bundle for obj: '%r' and with data: '%r'>"
+        if six.PY2:
+            repr_string = repr_string.encode('utf-8')
+        return repr_string % (self.obj, self.data)
--- a/tests/basic/tests/resources.py
+++ b/tests/basic/tests/resources.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from django.contrib.auth.models import User
 from django.http import HttpRequest
 from tastypie.fields import ToOneField, ToManyField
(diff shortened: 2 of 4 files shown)


tastypie/bundle.py
```diff
@@ -32,4 +33,7 @@ def __init__(self,
         self.via_uri = via_uri
 
     def __repr__(self):
-        return "<Bundle for obj: '%s' and with data: '%s'>" % (self.obj, self.data)
+        repr_string = "<Bundle for obj: '%s' and with data: '%s'>"
```
SeanHayes: Could you make this a unicode string (u""), perform the substitution, and then perform encode('utf-8') if six.PY2? Otherwise, if self.obj or self.data are unicode, the returned string could be unicode as well. It's unlikely, but better to be safe.

tests/basic/tests/resources.py
```diff
@@ -178,3 +178,14 @@ def test_rollback(self):
...
         self.assertRaises(SlugBasedNote.DoesNotExist, SlugBasedNote.objects.get, pk='first-post')
+
+
+class BundleTestCase(TestCaseWithFixture):
+    def test_bundle_repr(self):
+        # __repr__ should return string type (str in PY2 or unicode in PY3)
+        n1 = Note.objects.get(pk=1)
+
+        resource1 = NoteWithAnnotationsResource()
+        n1_bundle = resource1.build_bundle(obj=n1)
```
SeanHayes: Could you also pass in a data dict, and make one of its keys unicode?

tastypie/bundle.py
prokaktus: It's already unicode, because we're importing `unicode_literals`.

If we try to perform substitution first, we can get `UnicodeDecodeError`, because self.data or self.obj can contain non-ascii `str` (Django's models `repr` return `str` in Py2, converted from `__unicode__`-method). Maybe I should explicit call `repr` on `self.obj` and `self.data`?

SeanHayes: Ahh, I forgot about unicode_literals.

I just realized, probably the best solution is to use %r instead of %s.

SeanHayes:
Don't forget to add this PR and related issue to the list in docs/release_notes/dev.rst

tastypie/bundle.py
prokaktus: Yeah, you're right! Thanks, I'll make changes.

SeanHayes:
Thanks!

Issue #1423, opened 2016-01-19
PATCHing resources with read only M2M relations fails.


fluffels:
# Versions

* Present in 0.13 with Django 1.9
* Absent in 0.12.2 with Django 1.8

# Steps to replicate

A test case has been submitted as a pull request. The names of Resources used in this test case are used below.

1. Create a resource (Equation) with an M2M relation to another (Constant).
1. The relation must be `full=True`.
1. Constant must have Authorization set to ReadOnlyAuthorization.
1. POST a new Constant.
1. POST a new Equation that contains a reference to a Constant.
1. Attempt to PATCH the Equation.

# Expected Behaviour

The PATCH succeeds as normal, returning HTTP 202 Accepted.

# Observed Behaviour

The PATCH fails, returning HTTP 401 Unauthorized.

Worse, the relation will be cleared due to https://github.com/django-tastypie/django-tastypie/blob/master/tastypie/resources.py#L2446

SeanHayes:
From the PR:
"401 is expected, but deleted data is a bug. The best way to fix this is probably to do away with related_mngr.clear(). As a workaround for now, set readonly=True on your field."

OOPMan:
@SeanHayes I'm not sure how you can say 401 is expected here because it really shouldn't be.

The situation that @fluffels has described is one in which you have a Resource A with M2M related Resource B. For Resource A you want to be able to change the instances of Resource B associated with an instance of Resource A whether it's a POST or a PATCH that creates/changes that link. 

What you don't want is for someone to be able to POST a Resource A instance that includes an inline creation for an instance of Resource B.

So, you configure Resource B to be read-only and you configure the relation between Resource A and Resource B to be modifiable.

This worked fine in the 0.12.x line and the 0.13.x line has introduced the buggy behaviour described by @fluffels. Your advice that he set the M2M field to readonly completely misses the point of the desired behaviour.

I imagine this is a side effect of the peculiar way that Tastypie seems to perform PATCHes.

flanaman:
So is there no way to allow relation assignment when the related resource is readonly? I want to assign relations but only to existing model instances and without updating any fields on those instances.

PR #1424, opened 2016-01-19
Test case for issue #1423

fluffels:


tests/settings_related_resource.py
```diff
@@ -0,0 +1,5 @@
+from settings import *  # flake8: noqa
```
SeanHayes: I think this file is just a copy of  tests/settings_related.py

fluffels:
@SeanHayes Removed that duplicate file. Not sure how I missed that.

tests/related_resource/api/resources.py
```diff
@@ -278,3 +280,24 @@ class Meta:
...
+        queryset = Constant.objects.all()
+        authenticaion = Authentication()
+        authorization = ReadOnlyAuthorization()
+
+
+class EquationResource(ModelResource):
+    description = fields.CharField()
+    constants = fields.ToManyField(
+        ConstantResource, 'constants', related_name='equations', null=True,
+        blank=True, full=True)
```
SeanHayes: If you set readonly=True here, the tests will pass. I think the 401 is expected behavior, however deleting the related items isn't, the error should be raised before any changes are made to the data.

SeanHayes:
401 is expected, but deleted data is a bug. The best way to fix this is probably to do away with `related_mngr.clear()`.

tests/related_resource/api/resources.py
fluffels: @SeanHayes Won't that stop me from modifying this relation in future PATCHes?

SeanHayes: That won't be possible since the related resource has ReadOnlyAuthorization. If you want ConstantResource to have ReadOnlyAuthorization, but for EquationResource.constancts to be changeable, you'll need to create another resource with a different Authorization that permits creating/updating.

    class ConstantResource(ModelResource):
        class Meta:
            always_return_data = True
            queryset = Constant.objects.all()
            authenticaion = Authentication()
            authorization = ReadOnlyAuthorization()


    class WritableConstantResource(ConstantResource):
        class Meta(ConstantResource.Meta):
            authorization = AMorePermissibleAuthorization()


    class EquationResource(ModelResource):
        description = fields.CharField()
        constants = fields.ToManyField(
            WritableConstantResource, 'constants', related_name='equations', null=True,
            blank=True, full=True)

PR #1428, opened 2016-01-31
Fix for "PATCH on ModelResource clears Resource's ImageField data"

ketanbhatt:


ketanbhatt:
@SeanHayes Tests failing for older versions because `_meta.get_fields()` was introduced in django 1.8.

Do I write an if condition for that? The methods that were in older versions are being deprecated. Please guide.

SeanHayes:
Yeah, just write an if statement for versions less than 1.8.

ketanbhatt:
@SeanHayes Tests unrelated to my branch are failing. They run fine on my local. Can you review?
Thanks

SeanHayes:
Failures were due to bug in python-mimeparse.

I'm wondering if this is the best solution, or if we should change FileFields to not call "url" and just return the path. This bug is caused by tastypie dehydrating the FileFields and using the URL as the input, which is failing, which makes me doubt doing a PUT with the same data would work. In every other case tastypie tries to output data data from a GET that can be sent back as a PUT without issue.

tastypie/resources.py
```diff
@@ -270,6 +272,18 @@ def wrapper(request, *args, **kwargs):
 
         return wrapper
 
+    def handle_patch_for_filefields(self, bundle, data_to_patch):
+        obj = bundle.obj
+        data_to_patch_copy = data_to_patch.copy()
+        fixed_data = {}
+
+        for field in obj._meta.local_fields:
```
SeanHayes: This code presumes the Django ORM is being used, which may not be the case for the Resource class.

tastypie/resources.py
```diff
@@ -270,6 +272,18 @@ def wrapper(request, *args, **kwargs):
 
         return wrapper
 
+    def handle_patch_for_filefields(self, bundle, data_to_patch):
+        obj = bundle.obj
+        data_to_patch_copy = data_to_patch.copy()
```
SeanHayes: No need to copy since fixed_data is being updated and returned.

tastypie/resources.py
```diff
@@ -270,6 +272,18 @@ def wrapper(request, *args, **kwargs):
...
+        fixed_data = {}
+
+        for field in obj._meta.local_fields:
+            if type(field) in [ImageField, FileField]:
+                fixed_data[field.name] = field._get_val_from_obj(obj).name
```
SeanHayes: Any excluded FileFields on a model will get leaked into fixed_data.

SeanHayes:
It would be nice to change the way tastypie FileFields hydrate data, so that URLs would get changed back to Django storage paths, but the Django storage system doesn't currently provide a way of turning a storage URL into a path.

The best solution might be to just return the path, but also add a second, readonly URLField with the file's URL. If synced back this field will get ignored by tastypie.

ketanbhatt:
@SeanHayes but `path` is not implemented for all storage backends. [Django docs]((https://docs.djangoproject.com/en/1.9/ref/files/storage/#django.core.files.storage.Storage.path)) say that:

> path(name)[source]¶
The local filesystem path where the file can be opened using Python’s standard open(). For storage systems that aren’t accessible from the local filesystem, this will raise NotImplementedError instead.

So this will create problem if the storage backend is not based on the local filesystem, like `S3BotoStorage`.

ketanbhatt:
And to tackle turning URLs to Path, this is the hack I am using:
```python
urlparse(player.profile_pic.url).path
```

SeanHayes:
By "path" I mean the path within the Storage class (what gets stored in the DB), not the file system path, which may not be available. I would never recommend exposing file system paths.

ketanbhatt:
Oh okay. So yes. The `name` property. 
Would you do it, or should I try to fix it? I am a beginner so would take me more time :D

ketanbhatt:
@SeanHayes I just saw this PR.
Is this still happening, do you want me to fix ?

SeanHayes:
Up to you, but someone else will have to review it, I don't have the availability I used to.

georgedorn:
Merge request superseded by https://github.com/django-tastypie/django-tastypie/pull/1591

Issue #1383, opened 2015-10-14
Replace references to apply_authorization_limits and apply_limits from cookbook and code


SeanHayes:

mattbriancon:
@SeanHayes mind filling this one out a bit? Happy to fix docs, just not sure what needs fixing.

SeanHayes:
apply_authorization_limits appears a few times in the docs even though it's deprecated, possibly removed. We need to replace references to it in the docs with code that does the same thing, though I'm not sure what that is off the top of my head.

SeanHayes:
Duplicate of #1045

Issue #1045, opened 2013-10-08
apply_authorization_limits deprecated but still listed in documentation


chrismeyersfsu:
Using apply_authorization_limits it recommended as a way to enforce per-user-resources. However, the method is deprecated.

http://django-tastypie.readthedocs.org/en/latest/cookbook.html#creating-per-user-resources

Consider linking to the Authorization example that accomplishes per-user-resource enforcement.

http://django-tastypie.readthedocs.org/en/latest/authorization.html#implementing-your-own-authorization

prudnikov:
Totally agree with @honi comment regarding this issue. Any comments by core team on that?

SeanHayes:
Closed by #1433

Issue #837, opened 2013-02-25
Please update django-tastypie.readthedocs.org and tastypieapi.org


ivanvotti:
Documentation on [django-tastypie.readthedocs.org](https://django-tastypie.readthedocs.org) doesn't look consistent with the last version of tastypie.

For example [cookbook page](https://django-tastypie.readthedocs.org/en/latest/cookbook.html) still says: "This could be done by applying a filter in the `apply_authorization_limits` method of your resource."

But `apply_authorization_limits` has been completely deprecated (#809)

Documentation link on [http://tastypieapi.org/](http://tastypieapi.org/) is not correct. And current stable version is 0.9.12

SeanHayes:
Closed by #1433

Issue #275, opened 2011-10-26
Include imports for non-orm documentation


kylehayes:
It would be beneficial to have the imports at the top of the example on http://django-tastypie.readthedocs.org/en/latest/non_orm_data_sources.html#using-riak-for-messageresource

PR #1433, opened 2016-02-10
Cleaning up old, unneeded code.

SeanHayes:
TODO:
* [x] #1437
* [x] #1383 
* [x] #1045
* [x] #716 
* [x] #753 
* [x] #275
* [x] #770
* [x] #807
* [x] #810 
* [x] #837
* [x] #1442
* [x] #357

### DIFF
--- a/tastypie/resources.py
+++ b/tastypie/resources.py
@@ -61,11 +61,6 @@ def sanitize(text):
     return escape(text).replace('&#39;', "'").replace('&quot;', '"')
 
 
-class NOT_AVAILABLE:
-    def __str__(self):
-        return 'No such data is available.'
-
-
 class ResourceOptions(object):
     """
     A configuration class for ``Resource``.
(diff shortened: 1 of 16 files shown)


tastypie/resources.py
```diff
@@ -61,11 +61,6 @@ def sanitize(text):
     return escape(text).replace('&#39;', "'").replace('&quot;', '"')
 
 
-class NOT_AVAILABLE:
```
SeanHayes: Added in 075b2c4ca012374b67f8263376c63756d86c2ada, the one place in the code that used this was removed.

tastypie/resources.py
```diff
@@ -131,18 +126,15 @@ def __new__(cls, name, bases, attrs):
...
+        parents.reverse()
 
-            for p in parents:
-                parent_fields = getattr(p, 'base_fields', {})
+        for p in parents:
+            parent_fields = getattr(p, 'base_fields', {})
 
-                for field_name, field_object in parent_fields.items():
-                    attrs['base_fields'][field_name] = deepcopy(field_object)
-        except NameError:
```
SeanHayes: This code was part of the very first commit. This except block has zero code coverage, and I can't think of how a NameError would get raised here.

tastypie/resources.py
```diff
@@ -200,9 +192,10 @@ def __init__(self, api_name=None):
             self._meta.api_name = api_name
 
     def __getattr__(self, name):
-        if name in self.fields:
+        try:
```
SeanHayes: One fewer key lookup.

PR #1453, opened 2016-02-19
Updated Api class to build correct URLs even when api_name contains s…

tonnzor:
…pecial chars

If api_name contains special regex chars (e.g. "3.0") -- URLs generated would be not correct regular expressions.

Need to escape it first. So that for api_name="3.0" it would change from r"^(?P<api_name>%s)3.0/$" (INVALID!) to r"^(?P<api_name>3\.0)/$".

I've also tried another approach by passing regex into api_name, i.e. api_name=r"3\.0" but it broke some other things.

### DIFF
--- a/tastypie/api.py
+++ b/tastypie/api.py
@@ -1,4 +1,5 @@
 from __future__ import unicode_literals
+import re
 import warnings
 from django.conf.urls import url, include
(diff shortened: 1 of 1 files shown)


SeanHayes:
I'm guessing `Resource.base_urls()` has the same bug, we should fix it there too.

PR #1455, opened 2016-02-22
PK instead of URI serialization of RelatedResources

mikicz:
I love tastypie, it's much easier to implement than DRF, however, I use it with AngularJS, where api paths are handled by its $resource and I only need the primary key of the objects I'm getting - so I have to make an inherited custom foreign key where I override the serializing functionality to only return the primary key every time I want to use tastypie. I can't possibly be the only one who needs it, so I written a proper implementation of this functionality in this pull request. So if you set the  ``pk_only`` to True when defining the related field, you will receive only the PK of the objects instead of the URIs, and you will need to send only the pk when creating/updating objects.

I also written tests about it.


SeanHayes:
I really hate to turn down such a well written pull request, but I'm just not sure this belongs in Tastypie. It seems like a hack around the limitations of Angular and goes against the RESTful best practice of referring to resources by their URIs. Right now I'm really trying to clean up actual bugs and make Tastypie more standards compliant and more maintainable. Wouldn't adding compatibility code to your client be a better solution?

Tastypie also already supplies multiple ways to change data input and output:
1. Custom fields like you're using
2. Tell ToOneField to use a Resource you've defined that only has a 'pk' field defined
3. Use alter_deserialized_detail_data and alter_detail_data_to_serialize to rewrite {"pk": 1} to a URI and a URI back to a dict.

Soon I also plan on adding in a hook that will let you more easily choose a Field class to use when automatically adding introspected fields to a resource.

mikicz:
Shame, but yeah, I get what you're saying... Would it help if I took care of every bug that appears here regarding the functionality?

To the options:
2. But I use the other resources as well, with authorization, validation and other things, so I would be a pain to define a new Resource just with the pk.
3. That could work, but I don't like the speed implications. As far I saw in the code, there's a lot of code and processing around the resource uris and the pk serialization/deserialization is much more straightforward..

SeanHayes:
Thanks for understanding. :)

For #2, you could subclass your existing resource and its Meta class as well. Something like:

    class FooResource(ModelResource):
        class Meta:
            ...

    class PKFooResource(FooResource):
        class Meta(FooResource.Meta):
            fields = ['id']
            include_resource_uri = False

PR #1463, opened 2016-04-15
Update to Allow CleanedDataFormValidation to work with forms/models that have related fields.

andrekeller:
This adds tests for #923


andrekeller:
Hi,

this PR aims to fix #923, not sure if this is the right way to do this. I'm not sure if its OK to rebase this, as the code in validation.py is not mine.

Also I'm quite new to both testing and django-tastypie. Please be kind with your critics :-)

georgedorn:
Merged manually, thanks for the test.

PR #1476, opened 2016-07-22
JSONP security fix + tests

Fitblip:
Piggybacking on #1233's fix since I don't know if he still has a vested interest in trying to merge my tests in, but I just discovered this and it's quite concerning.

### DIFF
--- a/tastypie/utils/mime.py
+++ b/tastypie/utils/mime.py
@@ -28,7 +28,7 @@ def determine_format(request, serializer, default_format='application/json'):
             return serializer.get_mime_for_format(format)
 
     # If callback parameter is present, use JSONP.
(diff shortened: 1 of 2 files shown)


georgedorn:
Good catch and thanks for the tests.

Fitblip:
No problem! Thanks for the quick merge, a little disappointed that it took
~2 years to fix this given the severity of this vulnerability.

What sort of messaging will happen from here? I assume most folks running
TastyPie want to know that their APIs were potentially leaking information
like that.

When can we expect a release on pypi so we can upgrade?

georgedorn:
We're getting ready for a release for Django 1.10, so this will go out with that one.

PR #1479, opened 2016-08-02
Parse datetime consistency

olasfar:


tests/core/tests/fields.py
```diff
@@ -613,6 +613,12 @@ def test_model_resource_correct_association(self):
         api_field = ModelResource.api_field_from_django_field(models.DateTimeField())
         self.assertEqual(api_field, DateTimeField)
 
+    def test_convert_with_T(self):
+        self.assertEqual(DateTimeField().convert("2010-11-10T03:07:43"), aware_datetime(2010, 11, 10, 03, 7, 43))
```
mattbriancon: Leading 0 will be interpreted as octal. The tests will pass because `03 == 3` but the 0 should be dropped for clarity.

PR #1480, opened 2016-08-02
Django 1.10

georgedorn:
Adds compatibility for Django 1.10 while retaining backwards compatibility.

Issues discovered:
- Django 1.10 added a check during ModelUserBackend.is_authenticated() to disallow authentication by non-active users.  This makes tastypie's require_active flag no longer relevant while also making it basically impossible to allow non-active users to authenticate without rewriting Django's is_authenticated() method.  This invalidated some tests and also revealed some false-positive tests along the way.  Removing the tests for inactive user authentication caused coverage to drop slightly.
- Tastypie's Authenication.is_authenticated() is a bit of a mess, returning a truthy value (HttpUnauthorized) in some cases.  I feel like we should fix this to be a strictly True/False result and raise/return the HttpUnauthorized result at a higher level, which it may actually be doing already, but that's beyond the scope of this pull request.
- CSRF token handling changed slightly in Django 1.10, which broke the SessionAuthentication backend.  This required a fix via the compat module.

### DIFF
--- a/tests/core/tests/authentication.py
+++ b/tests/core/tests/authentication.py
@@ -33,8 +33,8 @@ class AuthenticationTestCase(TestCase):
         auth = Authentication()
         request = HttpRequest()
         # Doesn't matter. Always true.
-        self.assertTrue(auth.is_authenticated(None))
-        self.assertTrue(auth.is_authenticated(request))
+        self.assertEqual(auth.is_authenticated(None), True)
+        self.assertEqual(auth.is_authenticated(request), True)
 
     def test_get_identifier(self):
         auth = Authentication()
@@ -45,15 +45,6 @@ class AuthenticationTestCase(TestCase):
         request.META['REMOTE_ADDR'] = '127.0.0.1'
         request.META['REMOTE_HOST'] = 'nebula.local'
         self.assertEqual(auth.get_identifier(request), '127.0.0.1_nebula.local')
-
-    def test_check_active_false(self):
-        auth = Authentication(require_active=False)
-        user = User.objects.get(username='johndoe')
-        self.assertTrue(auth.check_active(user))
-
-        auth = Authentication(require_active=False)
-        user = User.objects.get(username='bobdoe')
-        self.assertTrue(auth.check_active(user))
 
     def test_check_active_true(self):
         auth = Authentication(require_active=True)
@@ -124,17 +115,9 @@ class BasicAuthenticationTestCase(TestCase):
         bob_doe.set_password('pass')
         bob_doe.save()
         request.META['HTTP_AUTHORIZATION'] = 'Basic %s' % base64.b64encode('bobdoe:pass'.encode('utf-8')).decode('utf-8')
-        self.assertFalse(auth.is_authenticated(request))
-
-    def test_check_active_false(self):
-        auth = BasicAuthentication(require_active=False)
-        request = HttpRequest()
-
-        bob_doe = User.objects.get(username='bobdoe')
-        bob_doe.set_password('pass')
-        bob_doe.save()
-        request.META['HTTP_AUTHORIZATION'] = 'Basic %s' % base64.b64encode('bobdoe:pass'.encode('utf-8')).decode('utf-8')
-        self.assertTrue(auth.is_authenticated(request))
+        auth_res = auth.is_authenticated(request)
+        # is_authenticated() returns HttpUnauthorized for inactive users in Django >= 1.10, False for < 1.10
+        self.assertTrue(auth_res is False or isinstance(auth_res, HttpUnauthorized))
 
 
 class ApiKeyAuthenticationTestCase(TestCase):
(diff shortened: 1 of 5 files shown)


tests/core/tests/authentication.py
```diff
@@ -33,8 +32,8 @@ def test_is_authenticated(self):
         auth = Authentication()
         request = HttpRequest()
         # Doesn't matter. Always true.
-        self.assertTrue(auth.is_authenticated(None))
-        self.assertTrue(auth.is_authenticated(request))
+        self.assertEqual(auth.is_authenticated(None), True)
```
georgedorn: auth.is_authenticated() sometimes returns HttpUnauthorized, which is a truthy value for assertTrue.  We should be very careful about using assertTrue for anything that doesn't return True/False.

We should probably consider fixing is_authenticated() because any is_* method should return only True or False, but it could break existing APIs.

tests/core/tests/authentication.py
```diff
@@ -124,17 +114,9 @@ def test_check_active_true(self):
         bob_doe.set_password('pass')
         bob_doe.save()
         request.META['HTTP_AUTHORIZATION'] = 'Basic %s' % base64.b64encode('bobdoe:pass'.encode('utf-8')).decode('utf-8')
-        self.assertFalse(auth.is_authenticated(request))
-
-    def test_check_active_false(self):
```
georgedorn: We might be able to do a @skipIf() here if we really wanted to support non-active user auth (which would only work in Django < 1.10).  I suggest we instead break from the past and remove the require_active flag from the Authentication class entirely; if a user wants to allow auth for a non-active user, they can do so explicitly by creating their own Authentication subclass.

georgedorn:
This is ready for a CR, and, once merged, a PyPI release.  I'm still not sure if I can do the latter.

Should we review outstanding PRs for inclusion in the PyPI release, or do so after?

tox.ini
```diff
@@ -16,17 +17,17 @@ test-executable =
...
-    dj{17,18,19,dev}: {[testenv]test-executable} test related_resource.tests --settings=settings_related
-    dj{17,18,19,dev}: {[testenv]test-executable} test alphanumeric.tests --settings=settings_alphanumeric
-    dj{17,18,19,dev}: {[testenv]test-executable} test authorization.tests --settings=settings_authorization
-    dj{17,18,19,dev}: {[testenv]test-executable} test content_gfk.tests --settings=settings_content_gfk
-    dj{17,18,19,dev}: {[testenv]test-executable} test customuser.tests --settings=settings_customuser
-    dj{17,18,19,dev}: {[testenv]test-executable} test namespaced.tests --settings=settings_namespaced
-    dj{17,18,19,dev}: {[testenv]test-executable} test slashless.tests --settings=settings_slashless
-    dj{17,18,19,dev}: {[testenv]test-executable} test validation.tests --settings=settings_validation
-    dj{17,18,19,dev}: {[testenv]test-executable} test gis.tests --settings=settings_gis_spatialite
+    dj{17,18,19,110,dev}: {[testenv]test-executable} test -p '*' core.tests --settings=settings_core
```
georgedorn: Shouldn't there be a better way to do dj{17,18,19,110,dev}?  That's all of the django versions.

SeanHayes: *Should* be, but I'm not sure if there is. Here's the relevant tox docs: http://tox.readthedocs.io/en/latest/config.html#complex-factor-conditions

georgedorn:
Any input before I merge this to master, probably Friday evening?  I'm making a couple design decisions and reducing coverage and I'd prefer not to do so unilaterally...

tests/core/tests/authentication.py
SeanHayes: I'd be in favor of that.

tests/core/tests/authentication.py
SeanHayes: Id' be in favor if making it return either True or False, it's a common point of confusion for new users. If we put it in a minor release instead of a micro release it should be fine.

SeanHayes:
You'll have to add this to `.travis.yml`:

    env:
      - DJANGO_VERSION=dj110

We'll probably see code coverage improve once that's done.

SeanHayes:
Looks like there's a test failure related to oauth_provider. If I recall correctly, that lib hasn't been updated in a long time and is a security risk. The best solution might be to drop support and let users roll their own OAuth using a newer library.

SeanHayes:
Maybe I'm misremembering, I found a recent version of the lib: https://pypi.python.org/pypi/django-oauth-plus/2.2.9

georgedorn:
Yeah, this error is utterly bizarre.  I don't get it locally, I'm not sure
what is actually installing the package, and it should be failing for more
than just Django 1.10.  I'm baffled.

tox.ini
```diff
@@ -41,14 +42,15 @@ deps =
...
+    dj110: Django>=1.10,<1.11
     djdev: https://github.com/django/django/archive/master.tar.gz
 
-    py27-dj{17,18,19}: django-oauth-plus==2.2.8
-    py27-dj{17,18,19,dev}: python-digest
-    py27-dj{17,18,19,dev}: oauth2
-    py27-dj{17,18,19,dev}: pysqlite==2.7.0
-    py34-dj{17,18,19,dev}: python3-digest>=1.8b4
-    dj{17,18,19,dev}: -r{toxinidir}/tests/requirements.txt
+    py27-dj{17,18,19,110}: django-oauth-plus==2.2.8
```
SeanHayes: I think those oauth tests only run if django-oauth-plus is installed, and before this line was changed we weren't installing it on dev, probably due to the failure. I'd try bumping the version or removing it for dj110.

georgedorn:
Finally, everything passes.  I'm going to merge to master and in the next few days we can start thinking about cutting a release.

PR #1482, opened 2016-08-08
Fixed bug in ToOneFields when attribute is not a string

bendemboski:
This addresses a regression introduced in af2c1d667f97a1a576e336766d7ac6245263b39d where a ToOneField with self.attribute set to None or a function would raise an exception during resource initialization.

### DIFF
--- a/tastypie/fields.py
+++ b/tastypie/fields.py
@@ -741,7 +741,7 @@ class ToOneField(RelatedField):
 
     def contribute_to_class(self, cls, name):
         super(ToOneField, self).contribute_to_class(cls, name)
-        if not self.related_name:
+        if not self.related_name and isinstance(self.attribute, six.string_types):
             related_field = getattr(self._resource._meta.object_class, self.attribute, None)
             if isinstance(related_field, ReverseOneToOneDescriptor):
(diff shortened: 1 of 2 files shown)


georgedorn:
Re-opening to re-run tests as the target builds have changed.

cristiano2lopes:
Thanks @bendemboski for this PR it's an example of a great PR. Test and fix. 
I cannot understand why is sitting here for so long. Is it because it's missing something ? Upstream update ?

georgedorn:
@cristiano2lopes It's a good PR, but it didn't pass tests on all supported versions of django/python at the time.  This usually suggests some level of compatability layer is needed, or simply that an assumption that works on one version of python or django doesn't apply to all of the supported ones.

vkuzma:
Is this PR going to be accepted any time soon?

georgedorn:
@vkuzma It needs to be rebased on the latest version and the failing tests need to be addressed.

georgedorn:
Rebased against `master` in #1583 where it is passing tests.  It can't be merged until whatever shuffle happened on travis and/or in flake8 is re-fixed.

PR #1483, opened 2016-08-17
removed slumber.in reference

kevinfree2013:
slumber.in website isn't functional hence removing it from docs

### DIFF
--- a/docs/tools.rst
+++ b/docs/tools.rst
@@ -43,14 +43,6 @@ Adds MsgPack_ support to Tastypie's serializer.
 Python
 ======
 
-Slumber
--------
-
-http://slumber.in/
-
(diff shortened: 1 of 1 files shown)


kevinfree2013:
my first pull request. please see if i need to do anything more.
thanks.

georgedorn:
The python library itself still exists even if the site does not.  While it has been a year since the last release, it's still a functioning library.  

We could update the docs point to the github page (https://github.com/samgiles/slumber) or the PyPI page (https://pypi.python.org/pypi/slumber/) or both, instead of deleting the reference entirely.

georgedorn:
Closing in favor of updating the url to github/pypi.

PR #1485, opened 2016-09-05
Correct 404 status text to "Not Found"

guitarmanvt:
Per RFC 2616, 404 is "Not Found". (410 is "Gone", which might arguably be better.) Tastypie seems to actually send the correct status text `404 Not Found`, so the problem was simply a documentation mismatch.


guitarmanvt:
Withdrawing PR due to lack of action from upstream.

PR #1486, opened 2016-09-14
Fix warnings

mattbriancon:
Fix a lot of warnings. Most are from tests but fixes a bunch that occur when using TP too.


mattbriancon:
There's a test failing in the py35-dj17 build but I'd like to suggest dropping support for Django 1.7 as we're almost a year past the extended support period.

georgedorn:
I'm in favor of dropping support for Django 1.7 as well; going from 1.7 to 1.8 is not a lot of work (compared with, say, 1.5 to 1.6).  If you want to add a commit to drop 1.7 from the travis tests, we can get this merged.

mattbriancon:
@georgedorn updated! Coveralls is complaining but Travis is :ship:

mattbriancon:
@SeanHayes any thoughts on getting this merged?

PR #1469, opened 2016-06-08
Advanced Filtering example missed ignore_bad_filters paramter

whitekid:

PR #1499, opened 2017-02-14
update documentation of Advanced Filtering

eshellman:
At some point, a keyword argument "ignore_bad_filters" was added to the build_filters method. This broke custom resources that implemented build_filters with the signature given in the documentation. I have updated the documentation to reflect the possible addition of keyword arguments to build_filters.


eshellman:
duplicates #1469

PR #1507, opened 2017-05-16
Fix Django 2.0 deprecation warnings

dannyroberts:
In Django 1.9 and before, `on_delete` defaults to `models.CASCADE`, but in Django 2.0 it will be a required argument. This PR simply adds in the argument set to its default, as suggested by this deprecation warning:

```
[...]lib/python2.7/site-packages/tastypie/models.py:35: RemovedInDjango20Warning: on_delete will be a required arg for OneToOneField in Django 2.0. Set it to models.CASCADE on models and in existing migrations if you want to maintain the current default behavior. See https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.ForeignKey.on_delete
  user = models.OneToOneField(AUTH_USER_MODEL, related_name='api_key')
```

PR #1508, opened 2017-05-16
Fixed failing flake8 tests

dannyroberts:
I think this is all that was failing on https://github.com/django-tastypie/django-tastypie/pull/1507


dannyroberts:
✅

PR #1509, opened 2017-06-14
Django 1.11

georgedorn:
WIP; updating to Django 1.11.

- Remove test matrix for 1.7, 1.9; we should support the most recent version (1.11), the most recent LTE (1.8), and run tests against dev (but allowed to fail).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,10 +10,8 @@ env:
   - MODE=flake8
   - MODE=flake8-strict
   - MODE=docs
-  - DJANGO_VERSION=dj17
   - DJANGO_VERSION=dj18
-  - DJANGO_VERSION=dj19
-  - DJANGO_VERSION=dj110
+  - DJANGO_VERSION=dj111
   - DJANGO_VERSION=djdev
 
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ matrix:
       env: MODE=flake8
     - python: "3.4"
       env: MODE=flake8-strict
+    - python: "2.7"
+      env: DJANGO_VERSION=djdev
 
 addons:
   apt:
--- a/tests/core/models.py
+++ b/tests/core/models.py
@@ -1,12 +1,11 @@
 from itertools import count
 import uuid
 
-from django.contrib.auth import get_user_model
 from django.db import models
 
 from tastypie.utils import now, aware_datetime
+from tastypie.compat import AUTH_USER_MODEL
 
-User = get_user_model()
 
 class DateRecord(models.Model):
     date = models.DateField()
@@ -18,7 +17,7 @@ class DateRecord(models.Model):
 
 
 class Note(models.Model):
-    author = models.ForeignKey(User, related_name='notes', blank=True, null=True)
+    author = models.ForeignKey(AUTH_USER_MODEL, related_name='notes', blank=True, null=True)
     title = models.CharField("The Title", max_length=100)
     slug = models.SlugField()
     content = models.TextField(blank=True)
@@ -48,7 +47,7 @@ class Note(models.Model):
 
 
 class NoteWithEditor(Note):
-    editor = models.ForeignKey(User, related_name='notes_edited')
+    editor = models.ForeignKey(AUTH_USER_MODEL, related_name='notes_edited')
 
     class Meta:
         app_label = 'core'
--- a/tests/core/models.py
+++ b/tests/core/models.py
@@ -117,24 +117,20 @@ class MyDefaultPKModel(models.Model):
         app_label = 'core'
 
 
-if hasattr(models, 'UUIDField'):
-    class MyUUIDModel(models.Model):
-        id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
-        anotheruuid = models.UUIDField(default=uuid.uuid4)
-        content = models.TextField(blank=True, default='')
-        order = models.IntegerField(default=0, blank=True)
+class MyUUIDModel(models.Model):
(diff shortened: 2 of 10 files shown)


georgedorn:
@SeanHayes Mind doing a review?  This doesn't finish implementing 1.11 compatability (there's still warnings to squelch), but it does update the test suite and travis configs to run the right tests.

.travis.yml
```diff
@@ -25,6 +23,8 @@ matrix:
       env: MODE=flake8
     - python: "3.4"
       env: MODE=flake8-strict
+    - python: "2.7"
+      env: DJANGO_VERSION=djdev
```
georgedorn: Django 2.0 doesn't support python 2.7 anymore.

tests/core/models.py
```diff
@@ -1,20 +1,23 @@
...
 
 
 class DateRecord(models.Model):
     date = models.DateField()
     username = models.CharField(max_length=20)
     message = models.CharField(max_length=20)
 
+    class Meta:
+        app_label = 'core'
+
```
georgedorn: Setting 'app_label' fixes a problem where non-core tests were using these models without having them in INSTALLED_APPS.

tests/core/models.py
```diff
@@ -1,20 +1,23 @@
...
 
 class Note(models.Model):
-    author = models.ForeignKey(User, related_name='notes', blank=True, null=True)
+    author = models.ForeignKey(AUTH_USER_MODEL, related_name='notes', blank=True, null=True)
```
georgedorn: Since these models get used during the customuser tests, 1.11 was throwing an error because it noticed the manager was pointing to the wrong model.

tests/core/models.py
```diff
@@ -92,20 +113,24 @@ class MyDefaultPKModel(models.Model):
...
+    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)
+    anotheruuid = models.UUIDField(default=uuid.uuid4)
+    content = models.TextField(blank=True, default='')
+    order = models.IntegerField(default=0, blank=True)
+
+    class Meta:
+        ordering = ('order',)
+        app_label = 'core'
 
-if hasattr(models, 'UUIDField'):
```
georgedorn: Every version of Django that we support now has models.UUIDField, so this check is no longer needed.

tests/core/tests/authentication.py
```diff
@@ -355,6 +355,7 @@ def test_check_active_true(self):
 
 
 @skipIf(not oauth2 or not oauth_provider, "oauth provider not installed")
+@skipIf(settings.DJANGO_VERSION >= settings.DJANGO_11, 'oauth-plus not compatible with django 1.11')
```
georgedorn: oauth_provider / oauth-plus hasn't been updated in a while.  there are forks that have been, and we might need to point users at those or update the test runner to install one of those instead.  maybe fork one of those forks into the django-tastypie org to ensure it stays put?

tests/run_all_tests.sh
```diff
@@ -41,6 +41,7 @@ for pytestpath in $PYTESTPATHS; do
         spatialite tastypie-spatialite.db "SELECT InitSpatialMetaData();"
     fi
     
-    ./manage_$type.py test $test_name.tests --traceback
+    echo "./manage_$type.py test $test_name.tests --traceback -t $test_name"
+    ./manage_$type.py test $test_name.tests --traceback -t $test_name
```
georgedorn: This was the result of a very confusing bug, where the test discovery was picking up tests in the wrong subdirectories.  -t specifies the top-level directory to look in for tests.  A whole stackoverflow QA on the subject: https://stackoverflow.com/questions/44550805/how-to-disable-djangos-test-discovery/44575787

tests/settings.py
```diff
@@ -1,12 +1,24 @@
 import os
 import sys
 
+from distutils.version import StrictVersion
+import django
+django_version_string = django.get_version()
+if django_version_string.startswith('2.0.dev'):
+    DJANGO_VERSION = StrictVersion('2.0')
```
georgedorn: This is gross, but I needed a way to not run oauth tests if not supported by the current django version.  I'm open to other ideas.

tox.ini
```diff
@@ -17,17 +15,17 @@ test-executable =
...
-    dj{17,18,19,110,dev}: {[testenv]test-executable} test related_resource.tests --settings=settings_related
-    dj{17,18,19,110,dev}: {[testenv]test-executable} test alphanumeric.tests --settings=settings_alphanumeric
-    dj{17,18,19,110,dev}: {[testenv]test-executable} test authorization.tests --settings=settings_authorization
-    dj{17,18,19,110,dev}: {[testenv]test-executable} test content_gfk.tests --settings=settings_content_gfk
-    dj{17,18,19,110,dev}: {[testenv]test-executable} test customuser.tests --settings=settings_customuser
-    dj{17,18,19,110,dev}: {[testenv]test-executable} test namespaced.tests --settings=settings_namespaced
-    dj{17,18,19,110,dev}: {[testenv]test-executable} test slashless.tests --settings=settings_slashless
-    dj{17,18,19,110,dev}: {[testenv]test-executable} test validation.tests --settings=settings_validation
-    dj{17,18,19,110,dev}: {[testenv]test-executable} test gis.tests --settings=settings_gis_spatialite
+    dj{18,111,dev}: {[testenv]test-executable} test -p '*' core.tests --settings=settings_core
```
georgedorn: Inexplicably, the `-t core` option isn't needed for these tests, only those run locally.

We also can probably remove all of the manage_foo.py files and replace them with just manage.py and the `--settings=settings_foo` option, both here and in run_all_tests.sh

PR #1510, opened 2017-06-19
Silence deprecation warnings

georgedorn:
- Makes warnings noisy during tests to make silencing them easier.
- Silences several deprecation warnings, as well as fixes some ancient warnings during tests.


georgedorn:
Continuing in #1512

PR #1512, opened 2017-06-26
Rebase fix warnings

georgedorn:
WIP, rebasing work from @mattbriancon and my own compat fixes to silence warnings and prepare for 2.0 compatibility.


georgedorn:
@mattbriancon I'm merging your warning fixes from #1486 with more fixes from adding 1.11 support.

I'm running into a new interesting problem in a test that exercises django session auth, which is failing on 1.11 but not on 1.8.  But when I manage to figure out what changed, this should all be ready to merge.

georgedorn:
Note to self:  the problem appears to be the `@csrf_exempt` decorator in `tastypie.resources.Resource.wrap_view` which now (django 1.11) prevents CSRF authorization entirely.

We might not need to test CSRF during session auth anyway; I'm not sure that the test is quite right to begin  with.

PR #1516, opened 2017-07-03
0.14.0 changelog: add note for removing ResourceTestCase

nikolas:
https://github.com/django-tastypie/django-tastypie/commit/09a29fdb80c6c6f8d4eb80e538ba1d4cd1e96732


georgedorn:
Ah, that's right.  I assumed that wasn't an end-user-facing change from the commit message alone.

PR #1519, opened 2017-07-10
Abstract resources

georgedorn:
Implement abstract ModelResources (partly to bypass `object_class` and `queryset` requirements.)

This came out of a prior change to more explicitly detect ModelResources lacking those attributes - the example cases were due to users creating abstract Resources and inheriting from them.


georgedorn:
This still needs documentation changes - abstract is a new flag.

PR #1526, opened 2017-08-29
fix doc: Http 200 OK is not HttpAccepted(202)

cheng10:


georgedorn:
Fixed in #1528, thanks for catching this.

PR #1531, opened 2017-10-23
Standardized auth identifiers 

### LINKED ISSUE #1465: Standardize auth identifiers
Right now `get_identifier` will return the ip:host of the current request, other times it will return the current username or 'nouser' if the request isn't authenticated.

This should be standardized to always be the username or the ip:host.

bigblind:
Fixes #1465, thanks @SeanHayes for bringing this up.


georgedorn:
This is a fine approach, but needs a couple minor changes:
- You've accidentally included `.coverage` which is essentially a scratch file.  We should add that to `.gitignore` to prevent this.
- Tests are needed, to show this works consistently and prevent future regressions.  I may take a stab at those myself in a week or two, unless you're interested; tests are normally fairly easy to implement, but this is one area where Django's test suite can be a little tricky.

Issue #1540, opened 2018-01-15
Deal with CI flake8 errors: 'do not use bare except'


pawelswiecki:
Travis CI is using the newest flake8 version, that is 3.5.0. This version stared checking for bare exceptions (i.e. catch-all exceptions). See this PR https://github.com/django-tastypie/django-tastypie/pull/1539. The following errors were reported byt flake8 CI step:

```
./tastypie/authentication.py:79:9: E722 do not use bare except'
./tastypie/resources.py:2266:13: E722 do not use bare except'
./tests/core/tests/resources.py:852:9: E722 do not use bare except'
./tests/core/tests/resources.py:4550:9: E722 do not use bare except'
```

To make Travis happy and to make the code better this needs to be dealt with.

georgedorn:
Marking this fixed by #1541

PR #1541, opened 2018-01-16
Deal with bare except statements. See #1540.

pawelswiecki:
Notes about the changes:

* `tastypie.authentication.Authentication.get_authorization_data` change

    This is about re-raising the ValueError with a custom message in case of
    .split() returning one item (instead of two). Made it explicit.

* `tastypie.resources.BaseModelResource.obj_update` change

    I left this as is, because pinpointing **all the possible exceptions** that
    may be raised during data hydration is error-prone. "# flake8: noqa"
    prevents flake8 error.

* `tests.core.tests.resources.ResourceTestCase.test_auth_check` change

    BasicResource.is_authenticated() (without a custom authentication defined
    in "meta") can only raise ImmediateHttpResponse exception. Catch it
    explicitly.

* `tests.core.tests.resources.ModelResourceTestCase.test_is_valid` change

    Current implementation of Resource.is_valid() won't -- in case of errors --
    raise an exception, but will return False. Changed the test to check for
    that.

### DIFF
--- a/tastypie/authentication.py
+++ b/tastypie/authentication.py
@@ -76,7 +76,7 @@ class Authentication(object):
 
         try:
             auth_type, data = authorization.split(' ', 1)
-        except:
+        except ValueError:
             raise ValueError('Authorization header must have a space separating auth_type and data.')
 
         if auth_type.lower() != self.auth_type:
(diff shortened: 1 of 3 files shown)


georgedorn:
Coverage change must be a rounding error.  Thanks for flake8 fixes.

Issue #1537, opened 2017-12-28
TypeError: 'NoneType' object is not iterable


SKisContent:
I'm hitting what appears to be a logic error in the `dehydrate` method of `ToManyField` class in _fields.py_:
The code in question starts at line 856: 
```
        if the_m2ms is None:
            if not self.null:
                raise ApiFieldError("The model '%r' has an empty attribute '%s' and doesn't allow a null value." % (previous_obj, attr))

        if isinstance(the_m2ms, models.Manager):
            the_m2ms = the_m2ms.all()

        m2m_dehydrated = [
            self.dehydrate_related(
                Bundle(obj=m2m, request=bundle.request),
                self.get_related_resource(m2m),
                for_list=for_list
            )
            for m2m in the_m2ms
        ]
```
Let's say `the_m2ms` is `None` and `self.null` is `False`. The code will not raise an error, will not execute the `if isinstance` clause and will land on the `for` iteration. At this point a `TypeError: 'NoneType' object is not iterable` is thrown.

To replicate this, create a `ModelResource` subclass  with a self-referencing related field:
```class MyResource(ModelResource):
    children = fields.ToManyField('self', 'children', null=True, full=True)
```

toto22450:
Hello @SKisContent,

I am currently facing the same issue. Did you find a solution ?

Moliholy:
I seems it's been introduced in version 0.14.0. Take a look:
https://github.com/django-tastypie/django-tastypie/blob/v0.13.3/tastypie/fields.py#L859
https://github.com/django-tastypie/django-tastypie/blob/v0.14.0/tastypie/fields.py#L859

The question is: why?

georgedorn:
> The question is: why?

Because there are no tests covering this situation, so a regression was inevitable.  If you could add a test in your PR that fails without your code change, I'll be happy to merge it and cut a new release (we need one for 'official' Django 2.0 compatibility anyway).

Without a test, this regression will almost certainly happen again.

SKisContent:
Hey, toto22450,
I never did. I think I just worked around it.

georgedorn:
@toto22450 @SKisContent 
I'm unable to replicate this issue in a unit test.  It appears to only be possible when using a callable attribute or a relationship specified with a string that returns `None`.  Can you post your `Resource` and the request that triggers the error?

Moliholy:
[...]

PR #1543, opened 2018-03-14
Raise an error if a model doesn't have an attribute

Moliholy:
This PR addresses the issue https://github.com/django-tastypie/django-tastypie/issues/1537

Description of the problem
====================

As explained in the issue, in version 0.14.0 this bug was introduced. I just restored the previous status of the code.

### DIFF
--- a/tastypie/fields.py
+++ b/tastypie/fields.py
@@ -856,6 +856,7 @@ class ToManyField(RelatedField):
         if the_m2ms is None:
             if not self.null:
                 raise ApiFieldError("The model '%r' has an empty attribute '%s' and doesn't allow a null value." % (previous_obj, attr))
+            raise ApiFieldError("The model '%r' does not have the attribute '%s'", (previous_obj, attr))
 
         if isinstance(the_m2ms, models.Manager):
             the_m2ms = the_m2ms.all()
--- a/tests/core/tests/fields.py
+++ b/tests/core/tests/fields.py
@@ -648,6 +648,19 @@ class UserResource(ModelResource):
         return '/api/v1/users/%s/' % bundle_or_obj.obj.id
 
 
+class NoteResource(ModelResource):
+    class Meta:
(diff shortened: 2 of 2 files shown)


Moliholy:
UPDATE:
=======

It turned out to happen when the model doesn't have an attribute, so what I did instead was to raise an error specifying the reason.

georgedorn:
@Moliholy From my own investigation, this bug can happen on non-model resources, too.  For example, if you provide an `attribute` lamda function that returns None, which is a valid value when `null=True`.

Moliholy:
Yes, this happened in my case because I had a very big model with lots of relationships between different entities and one of them turned out to have a bad attribute name by mistake. Before 0.14.0 I didn't notice because it wasn't an important attribute and nothing broke, but when making the migration to Django 1.11 and python3 I realized there was no way to make one of my own test pass. It took me a while to discover we were using a wrong attribute name, so my first approach was to simply return an empty array as before, but in my opinion it'd be better to raise an exception, mainly because that'd guide the developer in the right way and at the same time avoid situations where nothing breaks but nothing works.

I hope the explanation shows some light over this issue!

georgedorn:
Yeah, your situation should raise an exception, but the lambda situation shouldn't.  It sounds like we're missing a sanity check during `ToManyField.__init__()`.

georgedorn:
So there are no sanity checks during `ToManyField.__init__()`, but further testing reveals that an `ApiFieldError` is already raised if you specify a bad attribute by string.  Weirdly, it's valid to add an attribute with an invalid name that is always null; I don't know why you'd want to, but it also means you can POST to that attribute a list of random nonsense and it won't cause an error during hydration, so maybe there's a weird edge case people might use that for?

In any case, it looks like all of the major cases are covered by the existing code and the fix in #1544, so I can finally merge that one and cut a release.

Issue #1547, opened 2018-04-05
`get_multiple` shuld peform a single query to the database.


jorgeecardona:
Currently, the method `get_multiple` makes a single query per-object. A single query is possible. 

There is currently a PR in this direction that was never merged: https://github.com/django-tastypie/django-tastypie/pull/701 From this PR I don't like that new methods are added and cache as well is added. We could use the `obj_get_list` method first and then decided if we want to use its cached version.

Best, 
Jorge.

PR #1548, opened 2018-04-05
Get multiple in one query

jorgeecardona:
This solves Issue: https://github.com/django-tastypie/django-tastypie/issues/1547 in a different way that https://github.com/django-tastypie/django-tastypie/pull/701 since I use the already existing method obj_get_list to get a queryset and then filter that by the required ids.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -36,7 +36,7 @@ Contributors:
 * Ed Summers (edsu) for a setup.py patch.
 * Sébastien Fievet (zyegfryed) for the initial OAuth implementation.
 * Jacob Kaplan-Moss (jacobkm) for the PATCH patch.
-* jorgeecardona for a patch in renaming the ``objects`` name of the response.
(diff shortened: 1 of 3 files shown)


georgedorn:
Needs tests to prevent future regressions, and breaks an existing test that ensures the order of objects returned matches the order specified in the URL.  That seems like a relatively trivial nuance, but I guarantee somebody out there is relying on it.

jorgeecardona:
Hi, 

The code it is now returning the objects in the same order, the tests are passing and coverage was increased.

There is one detail that I would like to get someone to review. I use str() in a point to convert the id that are in the models (usually integers) to compare them with the strings coming from the pk_list. Since Python 2 and 3 handle strings and bytes in different ways I just want to check that the usage of str at that point is safe. Tests are passing so I guess is fine, but I am not aware of some mysterious way to have encoded (or not) fields as pk.

Best,
Jorge

georgedorn:
Thanks, we've been waiting for a good PR for this for a long time!

PR #1563, opened 2018-08-10
assertKeys -> use the standard assertItemsEqual

romuald:
Use standard assertItemsEqual() unittest method so that the difference
will be printed when an assertion fails


romuald:
Can't use same method with python 2/3

Issue #1561, opened 2018-08-01
ImportError: cannot import name 'QUERY_TERMS' on Django 2.1


bastbnl:
Hi,

I just updated to Django 2.1 from Django 2.0.7 using `django-tastypie==0.14.1` on python 3.5 and ran into this error:
```
...
  File "<frozen importlib._bootstrap_external>", line 673, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "/var/www/brownpapersession/dev/brownpapersession/brownpapersession/urls.py", line 16, in <module>
    from tastypie.api import Api, NamespacedApi
  File "/var/www/brownpapersession/dev/env/lib/python3.5/site-packages/tastypie/api.py", line 11, in <module>
    from tastypie.resources import Resource
  File "/var/www/brownpapersession/dev/env/lib/python3.5/site-packages/tastypie/resources.py", line 32, in <module>
    from django.db.models.sql.constants import QUERY_TERMS
ImportError: cannot import name 'QUERY_TERMS'

```

I'll see if I can find more information on `QUERY_TERMS`

lhopkins:
Looks like `QUERY_TERMS` [is removed](https://github.com/django/django/blob/a656a681272f8f3734b6eb38e9a88aa0d91806f1/docs/releases/2.1.txt#L396).

georgedorn:
This is turning into a fairly difficult problem to solve; tastypie was relying heavily on `QUERY_TERMS` to quickly pre-validate filters.  There doesn't appear to be a graceful way to regenerate that set of operators without overhauling `ModelResource.build_filters`.  This is long overdue, anyway; validating against a constant list of operators made it impossible to add more operators without monkey-patching, and the global list of operators meant it would allow nonsensical operations through for fields that make no sense, like a case-insensitive filter on an integer field.

The path forward is probably rebuilding `build_filters` to retrieve the Django field and use the new lookups api to validate the operation.

georgedorn:
This was fixed in #1564 and merged into the `django-2.1` branch.  That branch isn't ready for release due to other test failures (see #1562), but I've confirmed that the remaining test failures are unrelated to `QUERY_TERMS` so this issue can be closed.

georgedorn:
To avoid further confusion, I'll leave this issue open until the rest of 2.1 support is implemented and merged to master.

georgedorn:
[...]

PR #1564, opened 2018-08-13
Fix/django2.1 query terms

georgedorn:
Fix for Django 2.1 removing `QUERY_TERMS` which tastypie was relying on heavily for parsing filters.

Instead of a set of all possible operators, this change digs into the field in question, using Django's `get_lookups()` to retrieve those operators that work for the specific field.

When the field is a relation (ToMany, ToOne, etc) it inquires with the related field.  This only traverses one relation (no chaining), but TastyPie already refuses to support filters more than one layer deep.

This is a pretty big change to filtering; chances are good there are edge cases that aren't well-tested.  On the upside, this code wasn't great to begin with, and most people just use `ALL` or `ALL_WITH_RELATIONS`.

One small upside?  Custom operations will work now, if the user creates custom Django fields and specifies the lookups via `RegisterLookupMixin`.

### DIFF
--- a/tastypie/resources.py
+++ b/tastypie/resources.py
@@ -13,11 +13,12 @@ import django
 from django.conf import settings
 from django.conf.urls import url
 from django.core.exceptions import (
-    ObjectDoesNotExist, MultipleObjectsReturned, ValidationError,
+    ObjectDoesNotExist, MultipleObjectsReturned, ValidationError, FieldDoesNotExist
 )
 from django.core.signals import got_request_exception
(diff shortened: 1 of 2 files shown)


georgedorn:
Note this doesn't address all issues of running TastyPie on 2.1; I've validated that it has the exact same test failures as hacking in the old `QUERY_TERMS` as suggested here: https://github.com/django-tastypie/django-tastypie/issues/1561#issuecomment-409708793

PR #1571, opened 2018-09-01
Add filter aliases resource options

jaypatel11:
Context:
Tastypie resource allows filtering using django filters. There are scenarios where we would want to design a GET API with filters which are readable for the client and does not expose the underlying Django model. For eg. `v1/users/?posts__comments_created__gte=2015-10-21` can be simply expressed as  `v1/users/?comments_after=2015-10-21`. Right now, custom filters can be applied to such cases, but that would require overriding `build_filters` and `apply_filters`. This is where declaring an "alias" for filter like:
```
filter_aliases = {
 'comments_after' : 'posts__comments_created__gte'
}
```
would be handy


georgedorn:
This is a useful feature, but it'll need documentation before it can be released.  I'll try to roll it in for the next release; v0.14.2 was overdue for users trying to upgrade to django 2.1.

PR #1583, opened 2019-02-25
Port fix from #1482

georgedorn:
Rebased the changeset from #1482, a proposed fix for `contribute_to_clase` failing when `attribute` is not a string.


georgedorn:
Confirmed that the tests in #1482 fail without the extra check for string, and that the suggested fix works across all currently-supported versions.

PR #1588, opened 2019-04-08
Django 2.2 release

georgedorn:
Changes to get travis to build against Django 2.2 properly.  No library changes needed; Django 2.2 is already compatible.

Drops support for django 2.0, which left official support on 4/1/2019.


georgedorn:
The good news is that I suspect everything works for 2.2 without changes.

The bad news is django 2.2 requires sqlite3 v3.8.3+, and Travis can't be made to install this with Spatialite, yet.

I started going down the path of converting the travis tests to run against postgresql (which we should be doing anyway, given that it's a lot closer to real-world applications) but our tests are really not made for it, so it's going to be a lot of work.  That's being done in the `travis-postgres` branch.

Still to be done:
- add migrations for all models, as any FK to `auth.User` will fail without a proper migration.  They worked on sqlite3 because FK constraints are not enforced.
-- alternate option:  figure out how to get the auth.User migration to run before the automatically-created migrations for the models in the tests.  I'm not sure if this is even possible.
- A lot of tests assume sqlite3 behavior, as the PKs are reset every time the in-memory db is dumped and recreated; postgres doesn't do this, so we can't assume the first object created has a pk of 1, for example.

georgedorn:
Well this is interesting: https://travis-ci.org/django-tastypie/django-tastypie

Evidently we need migrations for every model in the test case, even on sqlite3 now, unless we work around it by disabling FK constraints in the setup.  I think this is a good enough reason to implement migrations and prep for either switching to testing on postgres, or better yet using the test matrix to cover postgres, sqlite and mysql all at once.

georgedorn:
Added some minor prep work to start implementing Django 3.0 support.  Mostly calculating the StrictVersion correctly, so that the errors we get from those tests are legitimate compatibility issues.

georgedorn:
The changes required to get travis to build (specifically adding repo keys on Xenial) should be considered unstable and will likely require update in the near future.  Travis appears to be undergoing some major changes, leading to a lot of clunky workarounds while they work out how to have both fast-building images and support for arbitrary aptitude packages...

PR #1590, opened 2019-05-16
Travis postgres

georgedorn:
WIP.

Django and/or Travis recently broke support for Spatialite on Xenial; rather than continue fighting to get that working, let's try moving the tests to PostGIS.  We can continue supporting Spatialite for local development via inheritable settings.


georgedorn:
We can revisit this in the future; for now, spatialite is working again.

PR #1591, opened 2019-06-10
Ketanbhatt kb 1018 fix change in filefield on patch

georgedorn:
Merge conflicts fixed from https://github.com/django-tastypie/django-tastypie/pull/1428

Attempting to address review items; specifically, decoupling `Resource` from the Django ORM.  Only `ModelResource` should assume the ORM is in use.


tastypie/resources.py
```diff
@@ -291,7 +293,17 @@ def get_response_class_for_exception(self, request, exception):
...
             return http.HttpBadRequest
 
-        return http.HttpApplicationError
+    def handle_patch_for_filefields(self, bundle, data_to_patch):
+        obj = bundle.obj
+        data_to_patch_copy = data_to_patch.copy()
+        fixed_data = {}
+
+        for field in obj._meta.local_fields:
+            if type(field) in [ImageField, FileField]:
```
georgedorn: `isinstance` is almost universally better than `type`.
```suggestion
            if isinstance(field, in [ImageField, FileField]):
```

tastypie/resources.py
```diff
@@ -291,7 +293,17 @@ def get_response_class_for_exception(self, request, exception):
         elif isinstance(exception, UnsupportedFormat):
             return http.HttpBadRequest
 
-        return http.HttpApplicationError
+    def handle_patch_for_filefields(self, bundle, data_to_patch):
+        obj = bundle.obj
+        data_to_patch_copy = data_to_patch.copy()
```
georgedorn: Entirely unneeded; we aren't modifying `data_to_patch` and we're returning a new dict.
```suggestion
```

tastypie/resources.py
```diff
@@ -17,6 +17,8 @@
 from django.core.signals import got_request_exception
 from django.core.exceptions import ImproperlyConfigured
 from django.db.models.fields.related import ForeignKey
+from django.db.models import ImageField, FileField
```
georgedorn: Unused import now.
```suggestion
from django.db.models import FileField
```

georgedorn:
Still need to figure out how to generalize this to non-ORM resources.  A more general approach might be:

#. Get the object in question
#. Marshall the object into a dict matching the incoming payload
#. Update the dict with the incoming values
#. Perform the write.

`FieldField` is tricky, but in principle this seems like a generalization that could work for all tricky fields.

tastypie/resources.py
```diff
@@ -291,7 +293,16 @@ def get_response_class_for_exception(self, request, exception):
...
-        return http.HttpApplicationError
+    def handle_patch_for_filefields(self, bundle, data_to_patch):
+        obj = bundle.obj
+        fixed_data = {}
+
+        for field in obj._meta.local_fields:
+            if isinstance(field, FileField):
+                fixed_data[field.name] = field._get_val_from_obj(obj).name
+
+        fixed_data.update(data_to_patch_copy)
```
georgedorn:
```suggestion
        fixed_data.update(data_to_patch)
```

tastypie/fields.py
```diff
@@ -226,10 +226,15 @@ class FileField(ApiField):
     dehydrated_type = 'string'
     help_text = 'A file URL as a string. Ex: "http://media.example.com/media/photos/my_photo.jpg"'
 
-    def convert(self, value):
+    def convert(self, value, for_update=False):
         if value is None:
             return None
 
+        if for_update:
+            # Use the internal value, not the url, when we're hydrating just to
```
georgedorn:
```suggestion
            # Use the internal value, not the url, when we're dehydrating just to
```

Issue #1593, opened 2019-07-10
Resource.dispatch() only allows `HttpResponse`


gbansaghi:
The `dispatch()` method in `tastypie.resources.Resource` performs an [`isinstance` check][isinstance] that I believe is too restrictive:

```py
# If what comes back isn't a ``HttpResponse``, assume that the
# request was accepted and that some action occurred. This also
# prevents Django from freaking out.
if not isinstance(response, HttpResponse):
    return http.HttpNoContent()
```

The issue is that `HttpResponse` is not the base class for _all_ other response classes. It for most of them (e.g. `JsonResponse`, `HttpResponseNotFound`, `HttpResponseForbidden` etc.), but notably `StreamingHttpResponse` inherits only from `HttpResponseBase` (and `FileResponse` inherits from `StreamingHttpResponse`). Thus, streaming or file responses will always generate a 204 No Content.

I believe the solution might be as simple as changing this (and maybe the two other checks in this file) to `isinstance(response, HttpResponseBase)`, and I'd be happy to create a pull request if that's all it is. I wanted to open an issue first, though, to get some feedback and see if doing this might break things.

[isinstance]: https://github.com/django-tastypie/django-tastypie/blob/master/tastypie/resources.py#L506

georgedorn:
Yeah, it seems likely that this was written prior to the existence of `HttpResponseBase`.  A PR for it would be helpful, especially if you could include a test of something like `FileResponse`.

PR #1595, opened 2019-07-12
change instance checks to HttpResponseBase

gbansaghi:
As mentioned in the [issue] I opened, `HttpResponse` is no longer the base class for all responses. 

- add a test to ensure that a `FileResponse` can be returned (fails without patch)
- update `resources.py` to check responses against `HttpResponseBase` (fixes broken test)

Tested with Python 3.5 and Django 2.1.

[issue]: https://github.com/django-tastypie/django-tastypie/issues/1593


gbansaghi:
I fixed the extra newline that caused the `flake8` failures. The contribution guidelines say:

> Squash your changes down [...] to one commit containing your failing tests and one more commit containing the fix that makes those tests pass.

Should I open a new PR with the fixed branch or is it okay to add one more commit to this PR?

georgedorn:
The guidelines were also written before github had automatic squashing of PRs.  I'll squash the whole PR when I merge it; add as many commits as you need.

And thanks!

Issue #1618, opened 2020-11-11
Breaking change to model resource filtering in version 14.3


dharvey-consbio:
### Summary

In versions following `14.1`, Tastypie model resources no longer support related field queries in request filters. The issue lies in how the following two methods interact:

1. `BaseModelResource.build_filters` (loops over filters in request)
2. `BaseModelResource.check_filtering` (recurses to validate filters and related models)

After thorough investigation I was able to isolate the specific problems and reimplement both methods in a way that fixed the problem for my project. I also created unit tests to test all filtering variations in order to prove less complex cases worked as they have before in previous releases.

My project is built on Python 3.8, and uses Django 2.2.

### Details

The problem arises with requests that contain filtering data like the following:
```python
url = reverse('api_dispatch_list', args=('v1', 'base_rsrc'))
data = {'related_rsrc__related_field__startswith': 'related text'}
```
The first issue I ran into was that  `BaseModelResource.build_filters` always appends `exact` to every filter provided, which changes the provided filter to `related_rsrc__related_field__startswith__exact`. This results in this error:
> 'exact' is not an allowed filter on the 'related_field' field.

When I added "exact" to the filtering for `related_field` in `related_rsrc`, then I got this:
> The 'related_field' field does not support relations.

I discovered this was because `BaseModelResource.build_filters` determines the query terms based on the first field (`related_rsrc`) and not the target field of that model (`related_field`).

In my case `related_field` was a `CharField`, which did support `startswith`, but because the recursive method `BaseModelResource.check_filtering` was not the one validating query terms, the `startswith` portion of the filter was being treated as another related resource.

### Solution

In the code below I have preserved much of the original code in order to minimize the diff.

```python
class ModelResource(ResourceMixin, resources.ModelResource):

    def build_filters(self, filters=None, ignore_bad_filters=False):
        """ Overrides BaseModelResource.build_filters to support query terms for related fields """

        if filters is None:
            filters = {}

        qs_filters = {}

        for filter_expr, value in filters.items():
            filter_bits = filter_expr.split(LOOKUP_SEP)
            field_name = filter_bits.pop(0)
            filter_type = self.resolve_filter_type(field_name, filter_bits, 'exact')

            if field_name not in self.fields:
                # It's not a field we know about. Move along citizen.
                continue

            try:
                lookup_bits = self.check_filtering(field_name, filter_type, filter_bits)
            except InvalidFilterError:
                if ignore_bad_filters:
                    continue
                else:
[...]

PR #1615, opened 2020-08-10
Check field lookups for build_filters recursively.

fatcop:

PR #1619, opened 2020-11-12
Fix for BaseModelResource.build_filters and check_filtering

dharvey-consbio:
### Summary

Addresses Issue #1618

In versions following `14.1`, Tastypie model resources no longer support related field queries in request filters. The issue lies in how the following two methods interact:

1. `BaseModelResource.build_filters` (loops over filters in request)
2. `BaseModelResource.check_filtering` (recurses to validate filters and related models)

### Details

The problem arises with requests that contain filtering data like the following:
```python
url = reverse('api_dispatch_list', args=('v1', 'base_rsrc'))
data = {'related_rsrc__related_field__startswith': 'related text'}
```
The first issue is that  `BaseModelResource.build_filters` always appends `exact` to every filter provided, which changes the provided filter to `related_rsrc__related_field__startswith__exact`. This results in this error:
> 'exact' is not an allowed filter on the 'related_field' field.

Adding "exact" to the filtering configuration for `related_field` in `related_rsrc`, still results in this:
> The 'related_field' field does not support relations.

This is because `BaseModelResource.build_filters` determines the query terms based on the first field (`related_rsrc`) and not the target field of that model (`related_field`). When `related_field` is a `CharField`, which supports `startswith`, because the recursive method `BaseModelResource.check_filtering` is not validating query terms at each level, the `startswith` portion of the filter is treated as another related resource.

### DIFF
--- a/tastypie/resources.py
+++ b/tastypie/resources.py
@@ -2007,7 +2007,7 @@ class BaseModelResource(Resource):
             raise InvalidFilterError("The '%s' field does not allow filtering." % field_name)
 
         # Check to see if it's an allowed lookup type.
-        if self._meta.filtering[field_name] not in (ALL, ALL_WITH_RELATIONS):
+        if filter_type != 'exact' and self._meta.filtering[field_name] not in (ALL, ALL_WITH_RELATIONS):
             # Must be an explicit whitelist.
             if filter_type not in self._meta.filtering[field_name]:
                 raise InvalidFilterError("'%s' is not an allowed filter on the '%s' field." % (filter_type, field_name))
@@ -2015,8 +2015,16 @@ class BaseModelResource(Resource):
         if self.fields[field_name].attribute is None:
             raise InvalidFilterError("The '%s' field has no 'attribute' for searching with." % field_name)
 
-        # Check to see if it's a relational lookup and if that's allowed.
-        if len(filter_bits):
+        if len(filter_bits) == 0:
+            # Only a field provided, match with provided filter type
+            return [self.fields[field_name].attribute] + [filter_type]
+
+        elif len(filter_bits) == 1 and filter_bits[0] in self.get_query_terms(field_name):
+            # Match with valid filter type (i.e. contains, startswith, Etc.)
+            return [self.fields[field_name].attribute] + filter_bits
+
+        else:
+            # Check to see if it's a relational lookup and if that's allowed.
             if not getattr(self.fields[field_name], 'is_related', False):
                 raise InvalidFilterError("The '%s' field does not support relations." % field_name)
 
@@ -2027,9 +2035,12 @@ class BaseModelResource(Resource):
             # if any. We should ensure that all along the way, we're allowed
             # to filter on that field by the related resource.
             related_resource = self.fields[field_name].get_related_resource(None)
-            return [self.fields[field_name].attribute] + related_resource.check_filtering(filter_bits[0], filter_type, filter_bits[1:])
 
-        return [self.fields[field_name].attribute]
+            next_field_name = filter_bits[0]
+            next_filter_bits = filter_bits[1:]
+            next_filter_type = related_resource.resolve_filter_type(next_field_name, next_filter_bits, filter_type)
+
+            return [self.fields[field_name].attribute] + related_resource.check_filtering(next_field_name, next_filter_type, next_filter_bits)
 
     def filter_value_to_python(self, value, field_name, filters, filter_expr,
             filter_type):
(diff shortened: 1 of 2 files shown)


dharvey-consbio:
I believe this fix is more thorough then the one in this PR:

https://github.com/django-tastypie/django-tastypie/pull/1615

dharvey-consbio:
There already seem to be tests to cover this in this file:

https://github.com/django-tastypie/django-tastypie/blob/master/tests/core/tests/resources.py#L2048-L2130

If these tests are indeed running and passing on the current build, please let me know. These are the cases that need coverage:

**Success Cases**
* `{'field1': 'val1'}` (single field query uses `exact`)
* `{'field1': 'val1', 'field2': 'val2'}` (multiple field query uses `exact`)
* `{'field1': 'val1', 'field2__contains': 'val2'}` (multiple field query with supported filter `contains`)
* `{'field1': 'val1', 'related_rsrc': 'slug'}` (related resource pk query uses `exact`)
* `{'field1': 'val1', 'related_rsrc__related_field': 'val3'}` (related field query uses `exact`)
* `{'field1': 'val1', 'related_rsrc__related_field__startswith': 'val3'}` (related field query with supported filter `startswith`)

**Failure Cases**
* `{'field1': 'val1', 'field2__icontains': 'val2'}` (multiple field query with unsupported filter `icontains`)
* `{'field1': 'val1', 'related_rsrc__related_field__istartswith': 'val3'}` (related field query with unsupported filter `istartswith`)

dharvey-consbio:
> I believe this fix is more thorough then the one in this PR: #1615

The reason I think so is twofold:
1. In #1615 the return value of `check_filtering` is changed to include `filter_type` (as part of a tuple)
2. In this changeset checking query terms is encapsulated in its own function for reuse

This changeset preserves all method signatures and return values.

dharvey-consbio:
The checks that were not successful were all on the `flake8` builds because of an import statement not at the top of the module in `tests.core.tests.__init__`. They are not related to this changeset.

dharvey-consbio:
There were already tests to cover this changeset (`tests.core.tests.resources.test_build_filters`):

https://github.com/django-tastypie/django-tastypie/blob/master/tests/core/tests/resources.py#L2048-L2130

These tests are indeed running and passing after some minor changes. These are the cases that require coverage:

**Success Cases**
* `{'field1': 'val1'}` (single field query uses `exact`)
* `{'field1': 'val1', 'field2': 'val2'}` (multiple field query uses `exact`)
* `{'field1': 'val1', 'field2__contains': 'val2'}` (multiple field query with supported filter `contains`)
* `{'field1': 'val1', 'related_rsrc': 'slug'}` (related resource pk query uses `exact`)
* `{'field1': 'val1', 'related_rsrc__related_field': 'val3'}` (related field query uses `exact`)
* `{'field1': 'val1', 'related_rsrc__related_field__startswith': 'val3'}` (related field query with supported filter `startswith`)

**Failure Cases**
* `{'field1': 'val1', 'field2__icontains': 'val2'}` (multiple field query with unsupported filter `icontains`)
* `{'field1': 'val1', 'related_rsrc__related_field__istartswith': 'val3'}` (related field query with unsupported filter `istartswith`)

dharvey-consbio:
The checks that were not successful were all on the `flake8` builds because of an import statement not at the top of the module in `tests.core.tests.__init__`. They are not related to this changeset.

tastypie/resources.py
```diff
@@ -2054,6 +2065,35 @@ def filter_value_to_python(self, value, field_name, filters, filter_expr,
...
+        if field_name not in self.fields:
+            raise InvalidFilterError("The '%s' field is not a valid field" % field_name)
+
+        try:
+            django_field_name = self.fields[field_name].attribute
+            django_field = self._meta.object_class._meta.get_field(django_field_name)
+            if hasattr(django_field, 'field'):
+                django_field = django_field.field  # related field
+        except FieldDoesNotExist:
+            raise InvalidFilterError("The '%s' field is not a valid field name" % field_name)
```
georgedorn, line 2080: I'd like to see a test for this case (and the other negative case on line 2032, unrelated to this patch).  I can still merge this and take a stab at those cases (and some others) before cutting the next release, when I next get some time.  The related fields code has had lots of regressions in the past from insufficient tests.

tastypie/resources.py
```diff
@@ -2054,6 +2065,35 @@ def filter_value_to_python(self, value, field_name, filters, filter_expr,
...
+
+    def resolve_filter_type(self, field_name, filter_bits, default_filter_type=None):
+        """ Helper to derive filter type from next segment in filter bits """
+
+        if not filter_bits:
+            # No filter type to resolve, use default
+            return default_filter_type
+        elif filter_bits[0] not in self.get_query_terms(field_name):
+            # Not valid, maybe related field, use default
+            return default_filter_type
```
igorgai, line 2092: @dharvey-consbio there is an issue hidden here: it is incorrect to return the `default_filter_type` here for all the cases. 

With the examples of several levels of nested relations like `post__author__in` or `post__author__name__startswith`, the requests will fail to return the expected result because the default value of `'exact'` will be used here. Instead, we should recursively descend the chain of the `filter_bits`, similarly to how it is done in the `check_filtering` method above, and use the proper "filter type" for the last field in the chain. See the example of the fix implementation here: https://github.com/Partender/django-tastypie/pull/1/files/

xiaolong3105:
您好，您的邮件我已收到，我会尽快处理。
 
Best Wishes!
晓龙

Issue #1620, opened 2021-02-24
ModelResource generates CharField for BigAutoField


TDaulat:
### Summary
We are converting our int(11) primary keys on our Django models to bigint(20).  After the conversion, the primary key fields in the tastypie resource body are now returned as string types.

### Details
The following line in the ModelResource class is the culprit:

https://github.com/django-tastypie/django-tastypie/blob/master/tastypie/resources.py#L1913

The following change is proposed:
```diff
- elif internal_type in ('IntegerField', 'PositiveIntegerField', 'PositiveSmallIntegerField', 'SmallIntegerField', 'AutoField', 'BigIntegerField'):
+ elif internal_type in ('IntegerField', 'PositiveIntegerField', 'PositiveSmallIntegerField', 'SmallIntegerField', 'AutoField', 'BigIntegerField','BigAutoField'):
```

PR #1621, opened 2021-02-24
Add support for BigAutoField in ModelResource

TDaulat:
Issue: https://github.com/django-tastypie/django-tastypie/issues/1620


georgedorn:
Verified via #1625

PR #1623, opened 2021-04-12
tests: Test against Django 3.1 and 3.2

jpmckinney:


jpmckinney:
Two authentication tests are failing. I did some noodling around with `SECURE_REFERRER_POLICY`, `DEFAULT_HASHING_ALGORITHM`, `SESSION_COOKIE_SAMESITE`, but not sure they are related to the solution. I don't see any relevant changes in `django.middleware.csrf`. I'm not sure where the issue is.

georgedorn:
Closed via #1624

Thanks @jpmckinney for the changesets.  The auth problem was yet another change to how csrf token salting/unsalting works in Django; it's an unstable API that we've had to have compatibility shims to handle for several versions now.

PR #1626, opened 2021-06-23
Remove usages of django.conf.urls in favor of django.urls.

tim-schilling:
Replaces all usages of django.conf.urls.url with
django.urls.re_path.

These usages have been deprecated since v2.2. They will be removed
in v4.0


tim-schilling:
I want to note this will actively break integrations with Django < 2.2. I know support for those versions was dropped, but I did spot some 1.8 backwards compatibility code.

georgedorn:
This did get added to a release at some point, but I think some conflicts had to be resolved manually so this was never marked closed.

Issue #1627, opened 2021-10-14
Save fails when a single resource is used with multiple other resources


blayzen-w:
I ran into an issue with saving certain types of data structures that contains a single resource that is used in two or more different to many relationships. The problem is difficult to explain but from what I can tell it boils down to ambiguity with the bundle.objects_saved or the create_identifier function with unsaved objects. This causes some items to be incorrectly marked as being already saved when they aren't, resulting in an exception being thrown in django `ValueError: save() prohibited to prevent data loss due to unsaved related object` and the resource returning a 500 error.

Take the following test models
```
class MyContainerModel(models.Model):
    name = models.CharField(max_length=128, blank=True, null=True)

    class Meta:
        app_label = 'core'


class MyContainerItemModel(models.Model):
    parent = models.ForeignKey(MyContainerModel, on_delete=models.CASCADE, related_name='item_set')
    name = models.CharField(max_length=128, blank=True, null=True)

    class Meta:
        app_label = 'core'


class MyContainerItemGroupingModel(models.Model):
    parent = models.ForeignKey(MyContainerModel, on_delete=models.CASCADE, related_name='item_grouping_set')
    grouping_item = models.ForeignKey(MyContainerItemModel, on_delete=models.CASCADE)

    class Meta:
        app_label = 'core'
```

With the following resources
```
class MyContainerItemModelResource(ModelResource):
    parent = fields.ForeignKey('core.tests.resources.MyContainerModelResource', 'parent')

    class Meta:
        queryset = MyContainerItemModel.objects.all()
        allowed_methods = ['get', 'put', 'post']
        authorization = Authorization()
        resource_name = 'my-container-item'


class MyContainerItemGroupingModel(ModelResource):
    parent = fields.ForeignKey('core.tests.resources.MyContainerModelResource', 'parent')
    grouping_item = fields.ForeignKey(MyContainerItemModelResource, 'grouping_item')

    class Meta:
        queryset = MyContainerItemGroupingModel.objects.all()
        allowed_methods = ['get', 'put', 'post']
        authorization = Authorization()
        resource_name = 'my-container-item-group'


class MyContainerModelResource(ModelResource):
    container_items = fields.ToManyField(MyContainerItemModelResource, 'item_set', blank=True)
    container_grouping_items = fields.ToManyField(MyContainerItemGroupingModel, 'item_grouping_set', blank=True)

    class Meta:
        queryset = MyContainerModel.objects.all()
[...]

PR #1628, opened 2021-10-14
Issue 1627 - Save fails when a single resource is used with multiple related resources

blayzen-w:
See issue [1627](https://github.com/django-tastypie/django-tastypie/issues/1627) for details


georgedorn:
Closing and re-opening to convince github actions to run the tests.

tastypie/resources.py
```diff
@@ -2406,6 +2406,7 @@ def save(self, bundle, skip_errors=False):
 
         if obj_id not in bundle.objects_saved or bundle.obj._state.adding:
             bundle.obj.save()
+            obj_id = self.create_identifier(bundle.obj)
```
georgedorn, line 2409: I don't understand how this one line makes a difference; the `obj_id` is already created on line 2405.

blayzen-w: When the first create_identifier function is called the object has not been saved and does not have a id. This makes the return value of that function always have None at the end (ex: `core.mycontaineritemmodel.None`) which is the same as every other unsaved object which causes the saving issue. Once the object is saved the create identifier function will return a unique value.

georgedorn: Ah, I see now; on the first call, the PK is `None`, so the `obj_id` ends up shadowing all other not-yet-saved objects.

georgedorn:
The fix makes sense now and the test correctly fails when the fix is not applied.  Style check is failing (see the flake8 step in any of the test matrix entries) but I'll merge once that's been fixed.

blayzen-w:
Thanks for taking a look at it. I'll start working on fixing the code style issues in a bit.

georgedorn:
Thanks for the PR; sorry I didn't get to it sooner, but the Travis CI debacle meant no test matrix system and it took several hours of effort to migrate away from it.  But I absolutely appreciate the fix + tests; it makes it a lot easier to be confident in the fix.

PR #1634, opened 2022-04-12
authorization docs: missed <id> for delete_detail

chinskiy:
Looks like missed <id> param
https://django-tastypie.readthedocs.io/en/latest/authorization.html#authorization-options


georgedorn:
Thanks for the catch!

PR #1636, opened 2022-05-31
Remove six, __future, and move to unittest.mock

ryanneilyoung:
I've done the work to move away from everything that uses the six library, I've removed all instances of __future as they are for python2, and I've moved to the built in unittest.mock. I've also ran the black formatter, and changed up some code to remove deprecation warnings from the test suite. It should be a fairly innocuous change altogether. 

I've split it into two commits, the first contains the bulk of the actual code changes, the second contains the formatting changes.


georgedorn:
I'm in favor of the re-styling with black, but it should be in a separate PR with an extra build step to enforce that style in the future.  That's harder than it sounds if there are lines that black can't auto-reformat, which is the main reason I haven't done it yet.

ryanneilyoung:
So I should try and pull out this reformatting stuff into a separate PR?

PR #1654, opened 2023-04-10
Combine fixes from two repositories

igorgai:
Combine fixes from two repositories:
1. https://github.com/Partender/django-tastypie/pull/1
2. ?


igorgai:
Not ready to be submitted here yet

PR #1658, opened 2023-06-07
Added view permission checks on DjangoAuthorization

rensieeee:
`DjangoAuthorization` docs are outdated, specifically on the `READ_PERM_CODE` definition. The link pointing to Django 1.9 no longer exists, but the latest link to Django 4.2 references the following:

[https://docs.djangoproject.com/en/4.2/topics/auth/default/#permissions-and-authorization](https://docs.djangoproject.com/en/4.2/topics/auth/default/#permissions-and-authorization)

```
Access to view objects is limited to users with the “view” or “change” permission for that type of object.
```

This PR allows `GET` calls by users with view permissions, but without change permissions. `PATCH/PUT` calls still require change permissions, and `GET` calls do still work with change permissions. Should be fully backwards compatible.

### DIFF
--- a/docs/authorization.rst
+++ b/docs/authorization.rst
@@ -68,7 +68,7 @@ The permissions required using ``DjangoAuthorization`` follow Django Admin's imp
 +=================================+==================+===============================================+======================+
 | ``POST <resource>/``            | create\_list     | ``add``                                       |                      |
 +---------------------------------+------------------+-----------------------------------------------+----------------------+
-| ``POST <resource>/<id>`` (\*)   | create\_detail   | ``add``                                       |                      |
+| ``POST <resource>/<id>``        | create\_detail   | ``add``                                       |                      |
 +---------------------------------+------------------+-----------------------------------------------+----------------------+
 | ``GET <resource>/``             | read\_list       | ``view``/``change``                           | ``update_list``      |
 +---------------------------------+------------------+-----------------------------------------------+----------------------+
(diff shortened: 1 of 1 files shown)


rensieeee:
In the meantime (temporary solution), you can implement something like this:

```
class CustomAuthorization(DjangoAuthorization):
    READ_PERM_CODE = 'view'
```

This is not backwards compatible (will restrict access to `GET` requests with change permissions). However, I'd argue that it's even better, because it restricts the influence and access to change permissions, which can still be overridden/fixed by giving the user the extra view permissions.

georgedorn:
Does anything change about TastyPie's authorization docs?  Or is this just a compatibility fix that was previously missed?

(Specifically, is https://github.com/django-tastypie/django-tastypie/blob/master/docs/authorization.rst still accurate or does it need updating?)

rensieeee:
First look seems to be that those docs need updating as well. 

[https://github.com/django-tastypie/django-tastypie/blob/master/docs/authorization.rst#djangoauthorization](https://github.com/django-tastypie/django-tastypie/blob/master/docs/authorization.rst#djangoauthorization)

This part in particular, but I'll read through the whole thing tomorrow. I'll update the PR with updates.

rensieeee:
First look seems to be that those docs need updating as well. 

[https://github.com/django-tastypie/django-tastypie/blob/master/docs/authorization.rst#djangoauthorization](https://github.com/django-tastypie/django-tastypie/blob/master/docs/authorization.rst#djangoauthorization)

This part in particular, but I'll read through the whole thing tomorrow. I'll update the PR with updates.

rensieeee:
I updated the docs, everything is up-to-date now.

Was thinking, maybe it's an idea to keep the `READ_PERM_CODE` implementation in another way? I guess a lot of implementations use this to exclude `GET` calls to view permissions, and this change will allow change permissions access as well. Maybe a check on this value to see if it's been set and if so, don't check on change permissions?

It will require a deprecation flag as well, but it'd be less breaking for existing implementations. But in this state it works too. Let me know what you think.

PR #1667, opened 2024-01-31
Use non-deprecated assertion methods

s-t-e-v-e-n-k:
Both TestCase.assert_() and TestCase.assertEquals() have been deprecated since Python 3.2, and removed as of Python 3.12. Switch to the correct method names, assertIn() and assertEqual().


georgedorn:
re-opening to see if that convinces github to re-run checks

georgedorn:
Merging into a rollup branch that'll be merged to master shortly for release.

PR #1671, opened 2024-07-19
 Fix race condition between POST / PATCH resources by using update_fields... #1151  Rebuilt against  v0.14.7.

georgedorn:
Manually rebasing @quietlyconfident's work in #1151 against the current release, in case anybody is interested in picking it back up.  I took an educated guess about how to merge that code into the current release; some of `BaseModelResource.save()` has already changed, but I cannot guarantee my guess is correct as I cannot replicate the issue.

Thus, the process for getting this merged:
1. Test this in a scenario known to cause the issue, to make sure my guess about merging is correct, and fix it if not.
2. Fix the current failing tests, making sure the new code doesn't introduce regressions.  It is possible the new code is the wrong way to do this.
3. Write new tests covering this new behavior; it would be nice if the race condition could be replicated via the live server test cases, but if it is a true race condition that may not be possible.

Again, I cannot take step 1 in this list, as I do not have a scenario known to cause the issue.

### DIFF
--- a/tastypie/resources.py
+++ b/tastypie/resources.py
@@ -1690,6 +1690,24 @@ class Resource(metaclass=DeclarativeMetaclass):
 
         # Now update the bundle in-place.
         deserialized = self.deserialize(request, request.body, format=request.META.get('CONTENT_TYPE', 'application/json'))
+        # Create a place to store the names of those fields we want to update
+        bundle.update_fields = [];
+        bundle.m2m_update_fields = [];
+        # When we get to the obj.save() stage, we need to know which fields have changed
+        # Otherwise we can't do a proper update.  Thus,
+        # For every key in deserialized (e.g. the fields submitted in the PATCH)
+        for key in deserialized:
+            # If the key is a property of the object, lets add it to the list, except:
+            if hasattr(bundle.obj,key):
+                # Can't update_fields an m2m field, so instead add it to patch_m2m_fields
+                if getattr(self.fields[key], 'is_m2m', False):
+                    bundle.m2m_update_fields.append(key)
+                    continue
+                # Don't add if it is the id/pk field, can't patch that.
+                if key == 'id' or key == 'pk':
+                    continue
+                # No more checks.  Add it.
+                bundle.update_fields.append(key)
         self.update_in_place(request, bundle, deserialized)
 
         if not self._meta.always_return_data:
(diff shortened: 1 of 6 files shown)


Poshi:
Thanks for your work here, @georgedorn :-)
Let's see if we can manage to make the proper code into the official release.
AFAIK, this seems a real race condition. The scenario you are lacking is simple, but it is not easy to generate for testing purposes, as it is a race condition. You just need two PATCH queries against the same object, but different fields. Ideally, the result would be as if both PATCH were applied sequentially, but when the race condition appears, only one of them will appear to be applied (although both would had been applied and the second PATCH to finish would overwrite the fields modified by the first PATCH with the original data). Not sure how could we cause a delay in one of the two PATCH in the proper location in the code to cause the race condition to appear predictably without modifying the code we are testing :-/

georgedorn:
To be clear, I am not certain that this patch will do anything to fix client-side race conditions.  I don't actually know if there is a way to sanely handle it in two requests, and the right answer might be to either force the requests to happen sequentially or do it all in one request.

Another option, if it can't be modeled at the http level, might be to use one or more Resource objects directly and simulate the order of events that cause the worst outcome.

Poshi:
I'm just a user that found the issue, but I'm not a specialist in tastypie/django. Never implemented anything with that framework, so this is just an opinion.

From my point of view, each PATCH should be translated to a SQL UPDATE query with only the relevant fields. It is quite clear to me that at some point (I don't know if this means tastypie or django), the **entire** object is retrieved, the fields are updated and the **entire** object is sent to the DB. That's the root cause of the race condition, but I don't know if this is tastypie implementation or django way-of-work. In case that the query needs to do the read-modify-write cycle for some reason, that set of SQL commands should be enclosed in a transaction with the proper isolation level (I think it should be serializable, but it's been a long time since I had to deal with the different isolation levels).

In any case, PATCH should make sure it only modifies the given fields, and now, for some reason, this is not being the case :-(

EndermanAPM:
Hi, I'm coming from the original #1151, @Poshi you're right it's trying to update the whole object when doing a PATCH, when it should only update the fields that are send in the requests.
What I could gather from the source, tastypie gets the requests, fetches the entire object from the database, updates the values using the requests and updates the entire object. (Using the same method as the PUT)
While debugging this issue I managed to create a test that would reproduce the issue and get fixed by this, I'll try to adapt it and make a PR.

georgedorn:
Ah, thanks, I'll cherry-pick those tests in to see if this fix works.

georgedorn:
This last set of test failures are unrelated; I'll work on fixing whatever broke in sphinx in a separate branch.

EndermanAPM:
Great! Happy to see we made progress. Hope to get the docs fixed soon so we can merge this.

EndermanAPM:
LGTM!

EndermanAPM:
@georgedorn whats the ETA for the new release?

PR #1696, opened 2025-12-12
tests/gis: don't rely on precise object IDs in database

Hyask:
Objects IDs can change, and are generally very brittle. That made the tests to fail when run in Debian autopkgtest for whatever reason. Testing coverage isn't actually impacted by removing those particular asserts on the "location", given that the location is used thereafter to re-query and verify the created object.

### DIFF
--- a/tests/gis/tests/views.py
+++ b/tests/gis/tests/views.py
@@ -161,10 +161,10 @@ class MoreViewsTestCase(TestCaseWithFixture):
         self.assertEqual(response.status_code, 201)
 
         location = response['Location']
-        self.assertTrue(location.endswith('/api/v1/geonotes/4/'))
+        self.assertIn('/api/v1/geonotes/', location)
 
         # make sure posted object exists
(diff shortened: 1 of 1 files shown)


tests/gis/tests/views.py
```diff
@@ -161,10 +161,9 @@ def test_post_object(self):
         self.assertEqual(response.status_code, 201)
 
         location = response['Location']
-        self.assertTrue(location.endswith('/api/v1/geonotes/4/'))
```
georgedorn, line 164: We should probably still at least check that the url we posted to is contained in the url we get back.

Hyask: Fair enough, updated.

tests/gis/tests/views.py
```diff
@@ -161,10 +161,10 @@ def test_post_object(self):
...
+        self.assertTrue('/api/v1/geonotes/' in location)
```
merwok: `assertIn` provides better failure message!

Hyask: Fixed, thanks for the feedback.

